PR #2, opened 2016-08-25
Initial drop of nodereport project for nodejs repo

rnchamberlain:
Implementation of native NPM module for diagnostic report triggered on exception and fatal error events, signals and JavaScript API. Supports Node v4 and v6, Linux, Mac and Windows. Includes testcases and demo samples.


Fishrock123:
I'm getting this on OS X 10.10.5 and Node.js v4/v6:

```
> node-gyp rebuild

  CXX(target) Release/obj.target/nodereport/src/node_report.o
../src/node_report.cc:675:76: warning: format specifies type 'long' but the argument has type '__darwin_suseconds_t' (aka 'int') [-Wformat]
    fprintf(fp,"\n  User mode CPU: %ld.%06ld secs", stats.ru_utime.tv_sec, stats.ru_utime.tv_usec);
                                       ~~~~~                               ^~~~~~~~~~~~~~~~~~~~~~
                                       %06d
../src/node_report.cc:676:78: warning: format specifies type 'long' but the argument has type '__darwin_suseconds_t' (aka 'int') [-Wformat]
    fprintf(fp,"\n  Kernel mode CPU: %ld.%06ld secs", stats.ru_stime.tv_sec, stats.ru_stime.tv_usec);
                                         ~~~~~                               ^~~~~~~~~~~~~~~~~~~~~~
                                         %06d
../src/node_report.cc:684:17: error: use of undeclared identifier 'RUSAGE_THREAD'
  if (getrusage(RUSAGE_THREAD, &stats) == 0) {
                ^
../src/node_report.cc:685:76: warning: format specifies type 'long' but the argument has type '__darwin_suseconds_t' (aka 'int') [-Wformat]
    fprintf(fp,"\n  User mode CPU: %ld.%06ld secs", stats.ru_utime.tv_sec, stats.ru_utime.tv_usec);
                                       ~~~~~                               ^~~~~~~~~~~~~~~~~~~~~~
                                       %06d
../src/node_report.cc:686:78: warning: format specifies type 'long' but the argument has type '__darwin_suseconds_t' (aka 'int') [-Wformat]
    fprintf(fp,"\n  Kernel mode CPU: %ld.%06ld secs", stats.ru_stime.tv_sec, stats.ru_stime.tv_usec);
                                         ~~~~~                               ^~~~~~~~~~~~~~~~~~~~~~
                                         %06d
4 warnings and 1 error generated.
make: *** [Release/obj.target/nodereport/src/node_report.o] Error 1
```

Also when applying this patch:
```
warning: squelched 20 whitespace errors
warning: 25 lines add whitespace errors.
```

mhdawson:
As an FYI Richard is on holiday this week , so response will be delayed.

mhdawson:
@bnoordhuis, believe you looked at PR to core, so you make be able to review this as the key issues will be the same, just as an npm versus part of core.

binding.gyp
```diff
@@ -0,0 +1,30 @@
+{
+  "targets": [
+    {
+      "target_name": "nodereport",
+      "sources": [ "src/node_report.cc", "src/module.cc" ],
+      "include_dirs": [ '<!(node -e "require(\'nan\')")' ],
+      "conditions": [
+        ["OS=='linux'", {
+          "defines": [ "_GNU_SOURCE" ],
+          "cflags": [ "-g -O2 -std=c++11", ],
```
bnoordhuis: These should be split, not as a single string.

binding.gyp
```diff
@@ -0,0 +1,30 @@
...
+      "dependencies" : [ "nodereport" ],
+      "copies": [
+        {
+          'destination': '<(module_root_dir)',
+          'files': ['<(module_root_dir)/build/Release/nodereport.node']
+        }]
+    },
+  ],
+}
+
```
bnoordhuis: Inconsistent use of single and double quotes in this file.

package.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "nodereport",
+  "main": "build/Release/nodereport.node",
+  "version": "0.1.0",
+  "keywords": [
+    ""
+  ],
```
bnoordhuis: Might as well leave out 'keywords' if you don't define any.

src/module.cc
```diff
@@ -0,0 +1,338 @@
...
+
+static v8::Isolate* node_isolate;
+
+/*******************************************************************************
+ * External JavaScript API for triggering a NodeReport
+ *
+ ******************************************************************************/
+NAN_METHOD(TriggerReport) {
+  Nan::HandleScope scope;
+  v8::Isolate* isolate = Isolate::GetCurrent();
```
bnoordhuis: You can use `info.GetIsolate()` instead of `Isolate::GetCurrent()`.  The former pulls it from the stack, the latter looks it up from the TLS.

src/module.cc
```diff
@@ -0,0 +1,338 @@
...
+
+  if (info[0]->IsString()) {
+    // Filename parameter supplied
+    Nan::Utf8String filename_parameter(info[0]->ToString());
+    if (filename_parameter.length() < 48) {
+      strcpy(filename, *filename_parameter);
+    } else {
+      Nan::ThrowSyntaxError("TriggerReport: filename parameter is too long (max 48 characters)");
+    }
+  }
```
bnoordhuis: Use a std::string so you don't have to impose a fairly arbitrary size limit?  At the very least compare against `sizeof(filename)` instead of a hard-coded value.

src/module.cc
```diff
@@ -0,0 +1,338 @@
...
+  if (info[0]->IsFunction()) {
+    // Callback parameter supplied
+    Nan::Callback callback(info[0].As<Function>());
+    // Creates a new Object on the V8 heap
+    Local<Object> obj = Object::New(isolate);
+    obj->Set(String::NewFromUtf8(isolate, "number"), Number::New(isolate, 54));
```
bnoordhuis: Why don't you use `Nan::New<...>` overloads for this?

src/module.cc
```diff
@@ -0,0 +1,338 @@
...
+NAN_METHOD(SetEvents) {
+  Nan::Utf8String parameter(info[0]->ToString());
+  v8::Isolate* isolate = node_isolate;
+  unsigned int previous_events = nodereport_events; // save previous settings
+  nodereport_events = ProcessNodeReportEvents(*parameter);
+
+  // If NodeReport newly requested for fatalerror, set up the V8 call-back
+  if ((nodereport_events & NR_FATALERROR) && !(previous_events & NR_FATALERROR)) {
+    isolate->SetFatalErrorHandler(OnFatalError);
+  }
```
bnoordhuis: Shouldn't you unset the handler if the flag is clear now but was set last time?

src/module.cc
```diff
@@ -0,0 +1,338 @@
...
+#ifndef _WIN32
+  // If NodeReport newly requested on external user signal set up watchdog thread and callbacks
+  if ((nodereport_events & NR_SIGNAL) && !(previous_events & NR_SIGNAL)) {
+    uv_sem_init(&report_semaphore, 0);
+    if (StartWatchdogThread(ReportSignalThreadMain) == 0) {
+      uv_async_init(uv_default_loop(), &nodereport_trigger_async, SignalDumpAsyncCallback);
+      uv_unref(reinterpret_cast<uv_handle_t*>(&nodereport_trigger_async));
+      RegisterSignalHandler(nodereport_signal, SignalDump, false);
+    }
+  }
```
bnoordhuis: You should check the return values of uv_sem_init() and uv_async_init() because those can fail (and sometimes will, i.e., it's not academic.)

src/module.cc
```diff
@@ -0,0 +1,338 @@
...
+  // Return value is the NodeReport filename
+  info.GetReturnValue().Set(Nan::New(filename).ToLocalChecked());
+}
+
+/*******************************************************************************
+ * External JavaScript APIs for nodereport configuration
+ *
+ ******************************************************************************/
+NAN_METHOD(SetEvents) {
+  Nan::Utf8String parameter(info[0]->ToString());
```
bnoordhuis: Calling ToString() is not necessary when the value is fed to `Nan::Utf8String`.

src/module.cc
```diff
@@ -0,0 +1,338 @@
...
+  }
+  // Trigger NodeReport if requested
+  if (nodereport_events & NR_FATALERROR) {
+    TriggerNodeReport(Isolate::GetCurrent(), kFatalError, message, location, NULL);
+  }
+  fflush(stderr);
+  if (nodereport_core) {
+    raise(SIGABRT); // core dump requested
+  } else {
+    exit(0); // no core dump requested
```
bnoordhuis: I don't think `exit(0)` is right for a fatal error.

src/module.cc
```diff
@@ -0,0 +1,338 @@
...
+
+#ifndef _WIN32
+static void SignalDumpInterruptCallback(Isolate *isolate, void *data) {
+  if (report_signal != 0) {
+    fprintf(stderr,"SignalDumpInterruptCallback - handling signal\n");
+    if (nodereport_events & NR_SIGNAL) {
+      if (nodereport_verbose) {
+        fprintf(stderr,"SignalDumpInterruptCallback - triggering NodeReport\n");
+      }
+      TriggerNodeReport(Isolate::GetCurrent(), kSignal_JS,
```
bnoordhuis: You get the isolate as a parameter, right?  No need to call `Isolate::GetCurrent()`.

src/module.cc
```diff
@@ -0,0 +1,338 @@
...
+    uv_sem_wait(&report_semaphore);
+    if (nodereport_verbose) {
+      fprintf(stderr, "Signal %s received by nodereport module\n", node::signo_string(report_signal));
+    }
+    uv_mutex_lock(&node_isolate_mutex);
+    if (auto isolate = node_isolate) {
+      // Request interrupt callback for running JavaScript code
+      isolate->RequestInterrupt(SignalDumpInterruptCallback, &report_signal);
+      // Event loop may be idle, so also request an async callback
+      size_t signo_data = static_cast<size_t>(report_signal);
```
bnoordhuis: I think passing (the address of) `report_signal` as a callback argument just obscures the fact that it's `report_signal`.  I'd drop it.

src/module.cc
```diff
@@ -0,0 +1,338 @@
...
+    uv_mutex_unlock(&node_isolate_mutex);
+  }
+  return nullptr;
+}
+
+static int StartWatchdogThread(void *(*thread_main) (void* unused)) {
+  pthread_attr_t attr;
+  pthread_attr_init(&attr);
+
+  pthread_attr_setstacksize(&attr, PTHREAD_STACK_MIN);
```
bnoordhuis: This won't work on freebsd, PTHREAD_STACK_MIN is so small you can't run anything on it.

src/module.cc
```diff
@@ -0,0 +1,338 @@
...
+    }
+    report_signal = 0;
+  }
+}
+
+static void RegisterSignalHandler(int signal, void (*handler)(int signal), bool reset_handler = false) {
+  struct sigaction sa;
+  memset(&sa, 0, sizeof(sa));
+  sa.sa_handler = handler;
+  sa.sa_flags = reset_handler ? SA_RESETHAND : 0;
```
bnoordhuis: SA_RESETHAND doesn't work with some versions of freebsd's libthr.  The workaround is to set the handler to SIG_DFL in the signal handler.

src/module.cc
```diff
@@ -0,0 +1,338 @@
...
+  if (report_name != NULL) {
+    ProcessNodeReportFileName(report_name);
+  }
+  const char* directory_name = getenv("NODEREPORT_DIRECTORY");
+  if (directory_name != NULL) {
+    ProcessNodeReportDirectory(directory_name);
+  }
+  const char* verbose_switch = getenv("NODEREPORT_VERBOSE");
+  if (verbose_switch != NULL) {
+	  nodereport_verbose = ProcessNodeReportVerboseSwitch(verbose_switch);
```
bnoordhuis: Indentation is off here.

src/module.cc
```diff
@@ -0,0 +1,338 @@
...
+  }
```
bnoordhuis: You should probably the getenv calls with a secure_getenv function that looks like this:
```c++
// Look up environment variable unless running as setuid root.
inline const char* secure_getenv(const char* key) {
#ifndef _WIN32
  if (getuid() != geteuid() || getgid() != getegid())
    return nullptr;
#endif
  return getenv(key);
}
```
(From node.cc.)

src/module.cc
```diff
@@ -0,0 +1,338 @@
...
+
+#ifndef _WIN32
+  // If NodeReport requested on external user signal set up watchdog thread and callbacks
+  if (nodereport_events & NR_SIGNAL) {
+    uv_sem_init(&report_semaphore, 0);
+    if (StartWatchdogThread(ReportSignalThreadMain) == 0) {
+      uv_async_init(uv_default_loop(), &nodereport_trigger_async, SignalDumpAsyncCallback);
+      uv_unref(reinterpret_cast<uv_handle_t*>(&nodereport_trigger_async));
+      RegisterSignalHandler(nodereport_signal, SignalDump, false);
+    }
```
bnoordhuis: This block of code is duplicated here: https://github.com/nodejs/nodereport/pull/2/files#r77156266

src/module.cc
```diff
@@ -0,0 +1,338 @@
...
+    fprintf(stdout, "Initialized nodereport module, event flags: %#x core flag: %#x\n",
+            nodereport_events, nodereport_core);
+#else
+    fprintf(stdout, "Initialized nodereport module, event flags: %#x core flag: %#x signal flag: %#x\n",
+            nodereport_events, nodereport_core, nodereport_signal);
+#endif
+  }
+}
+
+NODE_MODULE(nodereport, Initialize)
```
bnoordhuis: You should wrap the code in this file (and the header file) in a namespace.

src/node_report.cc
```diff
@@ -0,0 +1,790 @@
...
+  strcpy(report_filename, args);
+}
+
+void ProcessNodeReportDirectory(const char *args) {
+  if (strlen(args) == 0) {
+    fprintf(stderr, "Missing argument for nodereport directory option\n");
+    return;
+  }
+  if (strlen(args) > NR_MAXPATH) {
+    fprintf(stderr, "Supplied nodereport directory path too long (max 2048 characters)\n");
```
bnoordhuis: Use `"max %zu characters", NR_MAXPATH` here?  (Or better: `sizeof(report_directory) - 1`.)

src/node_report.cc
```diff
@@ -0,0 +1,790 @@
...
+    return;
+  }
+  strcpy(report_directory, args);
```
bnoordhuis: It's not exactly wrong but `strcpy()` will set off most static analyzers.  I'd use `snprintf(report_directory, sizeof(report_directory), "%s", args)`.

src/node_report.cc
```diff
@@ -0,0 +1,790 @@
...
+  fprintf(fp, "Module load time: %4d/%02d/%02d %02d:%02d:%02d\n",
+          loadtime_tm_struct.wYear, loadtime_tm_struct.wMonth, loadtime_tm_struct.wDay,
+		  loadtime_tm_struct.wHour, loadtime_tm_struct.wMinute, loadtime_tm_struct.wSecond);
+#else  // UNIX, OSX
+  fprintf(fp, "Dump event time:  %4d/%02d/%02d %02d:%02d:%02d\n",
+          tm_struct.tm_year+1900, tm_struct.tm_mon+1, tm_struct.tm_mday,
+          tm_struct.tm_hour, tm_struct.tm_min, tm_struct.tm_sec);
+  fprintf(fp, "Module load time: %4d/%02d/%02d %02d:%02d:%02d\n",
+		  loadtime_tm_struct.tm_year+1900, loadtime_tm_struct.tm_mon+1, loadtime_tm_struct.tm_mday,
+		  loadtime_tm_struct.tm_hour, loadtime_tm_struct.tm_min, loadtime_tm_struct.tm_sec);
```
bnoordhuis: Indent is off here.

src/node_report.cc
```diff
@@ -0,0 +1,790 @@
...
+#endif
+
+  // Print Node.js and deps component versions
+  fprintf(fp, "\nNode.js version: %s\n", NODE_VERSION);
+  fprintf(fp, "(v8: %s, libuv: %s, zlib: %s, ares: %s)\n",
+        V8::GetVersion(), uv_version_string(), ZLIB_VERSION, ARES_VERSION_STR);
```
bnoordhuis: And here.

src/node_report.cc
```diff
@@ -0,0 +1,790 @@
...
+  } else if (IsWindowsXPOrGreater()) {
+    fprintf(fp,"XP ");
+  }
+  if (IsWindowsServer()) {
+    fprintf(fp,"Server\n");
+  } else {
+    fprintf(fp,"Client\n");
+  }
+  TCHAR  infoBuf[256];
+  DWORD  bufCharCount = 256;
```
bnoordhuis: One space too many between the type and the variable name.  (Also, it's somewhat inconsistent to use camelCase here and snake_case everywhere else.)

src/node_report.cc
```diff
@@ -0,0 +1,790 @@
...
+  fflush(fp);
+#endif
+
+  // Print libuv handle summary (TODO: investigate failure on Windows)
+#ifndef _WIN32
+  fprintf(fp, "\n================================================================================");
+  fprintf(fp, "\n==== Node.js libuv Handle Summary ==============================================\n");
+  fprintf(fp,"\n(Flags: R=Ref, A=Active, I=Internal)\n");
+  fprintf(fp,"\nFlags Type     Address\n");
+  uv_print_all_handles(NULL, fp);
```
bnoordhuis: Be aware that `uv_print_all_handles()` "...is meant for ad hoc debugging, there is no API/ABI stability guarantees."  (From the documentation.)

src/node_report.cc
```diff
@@ -0,0 +1,790 @@
...
+  // Initialise the register state
+  state.pc = NULL;
+  state.fp = &state;
+  state.sp = &state;
+
+  isolate->GetStackSample(state, samples, arraysize(samples), &info);
+  if (static_cast<size_t>(info.vm_state) < arraysize(v8_states)) {
+      fprintf(fp, "JavaScript VM state: %s\n\n", v8_states[info.vm_state]);
+  } else {
+      fprintf(fp, "JavaScript VM state: <unknown>\n\n");
```
bnoordhuis: Indent is off here.

src/node_report.cc
```diff
@@ -0,0 +1,790 @@
...
+    return;
+  }
+  Local<StackTrace> stack = StackTrace::CurrentStackTrace(isolate, 255, StackTrace::kDetailed);
+  if (stack.IsEmpty()) {
+    fprintf(fp, "\nNo stack trace available from StackTrace::CurrentStackTrace()\n");
+    return;
+  }
+  // Print the stack trace, adding in the pc values from GetStackSample() if available
+  for (int i = 0; i < stack->GetFrameCount(); i++) {
+    if ((size_t)i < info.frames_count) {
```
bnoordhuis: Prefer `static_cast<size_t>(i)`.

src/node_report.cc
```diff
@@ -0,0 +1,790 @@
...
+    fprintf(fp,"\n  User mode CPU: %ld.%06ld secs", stats.ru_utime.tv_sec, stats.ru_utime.tv_usec);
+    fprintf(fp,"\n  Kernel mode CPU: %ld.%06ld secs", stats.ru_stime.tv_sec, stats.ru_stime.tv_usec);
+    fprintf(fp,"\n  Maximum resident set size: ");
+    WriteInteger(fp,stats.ru_maxrss * 1024);
+    fprintf(fp," bytes\n  Page faults: %ld (I/O required) %ld (no I/O required)", stats.ru_majflt, stats.ru_minflt);
+    fprintf(fp,"\n  Filesystem activity: %ld reads %ld writes", stats.ru_inblock, stats.ru_oublock);
+  }
+
+  fprintf(fp, "\n\nEvent loop thread resource usage:");
+  if (getrusage(RUSAGE_THREAD, &stats) == 0) {
```
bnoordhuis: Only some platforms have RUSAGE_THREAD.  Try wrapping it in an `#ifdef RUSAGE_THREAD`.

I'm not 100% sure if it's guaranteed to be a define, POSIX only requires that it's a "symbolic constant", but on linux and freebsd it's a define (and those are the only two that have RUSAGE_THREAD, I think.  OS X doesn't have it but you can query it through proc_info(), I think.)

src/node_report.cc
```diff
@@ -0,0 +1,790 @@
...
+  fprintf(fp, "\nEnvironment variables\n");
+  int index = 1;
+  char *env_var = *environ;
+
+  while (env_var != NULL) {
+    fprintf(fp, "  %s\n", env_var);
+    env_var = *(environ + index++);
+  }
+
+static struct {
```
bnoordhuis: You could make this `const`.

src/node_report.h
```diff
@@ -0,0 +1,38 @@
...
+#include "nan.h"
+
+using v8::Isolate;
+using v8::Local;
+using v8::Message;
+using v8::Function;
+using v8::Object;
+using v8::Number;
+using v8::String;
+using v8::Value;
```
bnoordhuis: You should really wrap this in a namespace.  `using` directives in the global namespace in header files are a perennial source of name conflicts.

rnchamberlain:
@Fishrock123 the two commits above should fix the compilation warnings + error you saw on OS X. I'll also fix the whitespace errors shortly, thanks.

Fishrock123:
Getting this now: (OS X 10.10.5)

```
> nodereport@0.1.0 install /Users/Jeremiah/Documents/nodesource/nodereport
> node-gyp rebuild

  CXX(target) Release/obj.target/nodereport/src/node_report.o
  CXX(target) Release/obj.target/nodereport/src/module.o
../src/module.cc:288:32: error: use of undeclared identifier 'secure_getenv'
  const char* verbose_switch = secure_getenv("NODEREPORT_VERBOSE");
                               ^
../src/module.cc:292:32: error: use of undeclared identifier 'secure_getenv'
  const char* trigger_events = secure_getenv("NODEREPORT_EVENTS");
                               ^
../src/module.cc:296:34: error: use of undeclared identifier 'secure_getenv'
  const char* core_dump_switch = secure_getenv("NODEREPORT_COREDUMP");
                                 ^
../src/module.cc:300:32: error: use of undeclared identifier 'secure_getenv'
  const char* trigger_signal = secure_getenv("NODEREPORT_SIGNAL");
                               ^
../src/module.cc:304:29: error: use of undeclared identifier 'secure_getenv'
  const char* report_name = secure_getenv("NODEREPORT_FILENAME");
                            ^
../src/module.cc:308:32: error: use of undeclared identifier 'secure_getenv'
  const char* directory_name = secure_getenv("NODEREPORT_DIRECTORY");
                               ^
6 errors generated.
make: *** [Release/obj.target/nodereport/src/module.o] Error 1
```

rnchamberlain:
@Fishrock123 darn, I really need a Mac to hand, plse try fix above

mhdawson:
@rnchamberlain you might try logging into our internal IBM OSX machines to do test builds there.

bnoordhuis:
@rnchamberlain You can just copy the secure_getenv implementation from node.cc, it's a glibc-ism otherwise.  Granted, the glibc version also checks capabilities, node does not.

rnchamberlain:
@bnoordhuis I found that nan.h pulled in the glibc secure_getenv implementation via stdlib.h and that gave a clash with a local one copied from node.cc. So I went for the official one on Linux, as it has a bit more functionality. But maybe some Linux installations would not have that level of glibc. I guess it is worth having the local implementation for OSX anyway. I'll have a think.

bnoordhuis:
> I found that nan.h pulled in the glibc secure_getenv implementation via stdlib.h and that gave a clash with a local one copied from node.cc.

And that is why you should use namespaces. :-)

rnchamberlain:
@bnoordhuis thanks for the comprehensive code review, I think I have covered the points now. 

A couple of notes:

1. Un-setting isolate->SetFatalErrorHandler(). There does not seem to be a way to 'unset' this
callback to revert to the default v8 behaviour. I tried re-setting it with a NULL pointer for the
callback, but the v8 behaviour is then a bit different (and less satisfactory) than the default, see code
in Utils::ReportApiFailure().

2. secure_getenv(). I'd rather avoid having a local implementation, so have gone with using the
glibc implementation if it is available, else simple getenv()

joshgav:
Just started reviewing :) <br><br>:octocat: *Sent from [GH](http://nodegh.io).*

mhdawson:
Is there a way that we can also provide command line options? Not necessary for the first pass but could be useful.

demo/exception.js
```diff
@@ -0,0 +1,34 @@
...
+   this.name = "UserException";
+}
+
+var http_server = http.createServer(my_listener);
+http_server.listen(8080);
+
+console.log('exception.js: Node running');
+console.log('exception.js: Go to http://<machine>:8080/ or http://localhost:8080/');
+
+setTimeout(function(){
```
mhdawson: Missing space before {

demo/loop.js
```diff
@@ -0,0 +1,55 @@
...
+    this.account = 98454324;
+}
+
+var http_server = http.createServer(my_listener);
+http_server.listen(8080);
+
+console.log('loop.js: Node running');
+console.log('loop.js: Go to http://<machine>:8080/ or http://localhost:8080/');
+
+setTimeout(function(){
```
mhdawson: missing space before {

src/node_report.cc
```diff
@@ -0,0 +1,785 @@
...
+#endif
+    break;
+  case kSignal_JS:
+  case kSignal_UV:
+    // Print the stack using StackTrace::StackTrace() and GetStackSample() APIs
+    PrintStackFromStackTrace(fp, isolate, event);
+    break;
+  }  // end switch(event)
+#endif
+  fflush(fp);
```
mhdawson: Should the code above be a separate function like the ones below.

package.json
```diff
@@ -0,0 +1,18 @@
...
+    "url": "https://github.com/nodejs/nodereport.git"
+  },
+  "engines": {
+    "node": ">=4.0.0"
+  },
+  "dependencies": {
+    "nan": "^2.3.5"
+  },
+  "license": "MIT"
+}
```
mhdawson: We should add the scripts entry so that the tests are run when you type 
npm test

mhdawson:
LGTM with a few nits.

rnchamberlain:
@mhdawson thanks for the review, I have moved the blocks of code that print version information and stack traces in node_report.cc into separate functions, and fixed the minor nits. Also added the scripts entry into package.json so uses can run "npm test", and fixed some CRLF line endings in the test code.

I had not considered line-mode options (options are currently supported via env vars and/or JS API calls). A typical node command line is:
```
node <node and v8 options> application.js <application options>
```
We could allow nodereport options to be mixed in with either the node/v8 or the application options. Suitable unique naming would be need, e.g something like "nodereport_events=exception+signal". Application options are available via process.argv, though I am not sure how to access that from C++, otherwise I should be able to get the command line from the OS . Anyway, something for a future enhancement.

Fishrock123:
Got it to build on OS X, is `NodeReport.20161004.120508.2622.001.txt` just meant to be read manually? Seems reasonably understandable to me.

Fishrock123:
Should this be merged so that it can be worked on more collaboratively?

rnchamberlain:
@Fishrock123 yes absolutely, the report is intended to be read manually, using any text editor/notepad/email etc, and should be readily understandable to node developers, admins, testers etc

Yes this is ready to go and for others to contribute. @mhdawson - not sure how this repo is set up, who has powers to merge this? thanks

Fishrock123:
I think we should probably just give you perms to merge it. cc/ @nodejs/post-mortem

misterdjules:
This repo has @mhdawson and the @nodejs/post-mortem team as collaborators. @rnchamberlain isn't (to my surprise) a member of the @nodejs/post-mortem team, even though he's clearly active in that space, so I added him to that team.

Once @rnchamberlain accepts the invite, he'll have the credentials to push commits to this repo.

README.md
```diff
@@ -8,3 +8,61 @@ It includes Javascript and native stack traces, heap statistics,
...
 reports are triggered on unhandled exceptions, fatal errors, signals
 and calls to a Javascript API.
+
+Usage:
+
+    npm install nodejs/nodereport
+
+    var nodereport = require('nodereport');
+
+By default, this will trigger a NodeReport to be written to the current
```
sam-github: I don't think this is a good idea. It means that there is no way to opt-out of auto-handling of `unhandledException`, which would be useful for someone who wanted to only use the manual `.triggerReport()`. I think it would make more sense for a require of `nodereport` to make no changes to global process state until it was attached. Syntax could be `require('nodereport')()`, or `require('nodereport').attach()`, or any other variant.

sam-github: also, I would want to control independently whether nodereport attached to SIGUSR2, or even configure the signal.

README.md
```diff
@@ -8,3 +8,61 @@ It includes Javascript and native stack traces, heap statistics,
...
+platform information section showing CPU and memory usage and system limits.
+The following messages are issued to stderr when a NodeReport is triggered:
+
+    Writing Node.js error report to file: NodeReport.201605113.145311.26249.001.txt
+    Node.js error report completed
+
+Configuration is available via these environment variables and/or Javascript
+API calls:
+
+    export NODEREPORT_EVENTS=exception+fatalerror+signal
```
sam-github: that its possible to configure nodereport via env vars is great, that its necessary to do so is I think too much focus on policy instead of mechanism

binding.gyp
sam-github: ^-- comment not addressed. js uses 4 SPC indents, c++ uses 2? I suggest pulling the eslint config from nodejs to keep some consistency of code, and make it easier for people to contribute consistent code to the project.

Fishrock123:
I don't think we should pull from core eslint. Just use `standard`.

binding.gyp
cjihrig: +1 to using the ESLint config from core.

README.md
rnchamberlain: Hi @sam-github, you can override the defaults and configure which events you want via the API, so
`require("nodereport").setEvents("fatalerror+signal");`
will switch off the exception handler, likewise
`require("nodereport").setEvents("exception+fatalerror");`
will switch off the signal handler.

`require("nodereport").setSignal("SIGQUIT");`
will configure SIGQUIT for the trigger signal instead of SIGUSR2, and we could add more choices for that (though I just found a bug in that code, fix on it's way).

README.md
rnchamberlain: I'd expect the `.setEvents()` API to be the usual way to configure it, with the env vars as an alternative maybe for admins when changing the code is not convenient. I'll update the readme to reflect that. 

Re the defaults, the design was for all the events to work 'out of the box' i.e. with just require("nodereport"). We could take a more cautious approach as you suggest, and just have 
`.triggerReport()` available until  `.setEvents(list of selected events)` is called.

mhdawson:
I believe I have given everybody in the post-mortem team write access to this repo. Was just waiting for 1 more L.G.T.M from one of the other reviewers before landing.

README.md
joshgav: Perhaps an optional options arg with key `events` as an array of potential triggers and key `sigusr2` as a boolean to enable/disable the signal would be a nice approach.

Such an arg would be reasonably extensible for other triggers too.

Also allows for clear defaults to be set early in code.

src/module.cc
```diff
@@ -0,0 +1,369 @@
...
+#ifndef _WIN32
+  // If NodeReport newly requested on external user signal set up watchdog thread and callbacks
+  if ((nodereport_events & NR_SIGNAL) && !(previous_events & NR_SIGNAL)) {
+    SetupSignalHandler();
+  }
+#endif
+}
+NAN_METHOD(SetCoreDump) {
+  Nan::Utf8String parameter(info[0]);
+  nodereport_events = ProcessNodeReportCoreSwitch(*parameter);
```
joshgav: I believe this is intended: `s/nodereport_events/nodereport_core`

rnchamberlain: @joshgav good spot, and same error in SetSignal method below, fix in progress

Fishrock123:
Status? I think I'd actually like to start using this.

rnchamberlain:
Fixed as discussed with @sam-github and @joshgav. So `require('nodereport')` by default now just enables the `.triggerReport()` functionality. No changes to the Node/v8 hooks and signals until the requirer explicitly calls  `.setEvents("exception+signal+apicall+etc")`.

Status: as per @mhdawson above, just another LGTM then we land. Meantime available to try out now with
 `npm install rnchamberlain/nodereport-1`

sam-github:
@bnoordhuis LGTY?

@rnchamberlain API LGTM. If you want to make it easy to enable all capabilities, you could implement a usage like:

```
require('nodereport')();
```

So a side-effectless require is possible, but all side-effects can be opted in easily, and the "module call" above would check for env vars

You could also extend above with an options object rather than method calls for configuration:

```
require('nodereport')({
  events: 'signal',
  signal: 'SIGHUP',
  });
```

rnchamberlain:
@sam-github I had a look at both of your suggested syntax options, but couldn't immediately work out  how to code them on the C++ side! There is a mention of a two-argument form of the Init() function here: https://nodejs.org/api/addons.html, which has a module object passed in as well as the exports object? Anyway yes, potential future enhancement - thanks.

.gitignore
```diff
@@ -0,0 +1,38 @@
...
+*.app
+*.node
+
+# Project files
+*.project
+*.cproject
+
+# Build directories
+/build/
+/node_modules/
```
bnoordhuis: This file uses Windows line endings for some reason.

README.md
```diff
@@ -1,10 +1,66 @@
...
+    export NODEREPORT_COREDUMP=yes|no
+    export NODEREPORT_VERBOSE=yes|no
+
+Sample programs for triggering NodeReports are provided in the
+node_modules/nodereport/demo directory: 
+
+    api.js - NodeReport triggered by Javascript API call
+    exception.js - NodeReport triggered by unhandled exception
+    fatalerror.js - NodeReport triggered by fatal error on Javascript heap out of memory
+    loop.js - looping application, NodeReport triggered using kill -USR2 <pid>
```
bnoordhuis: There is some trailing whitespace in this file.

demo/loop.js
```diff
@@ -0,0 +1,53 @@
...
+var http_server = http.createServer(my_listener);
+http_server.listen(8080);
+
+console.log('loop.js: Node running');
+console.log('loop.js: Go to http://<machine>:8080/ or http://localhost:8080/');
+
+setTimeout(function() {
+  console.log('loop.js: timeout expired, exiting.');
+  process.exit(0); 
+}, 60000);
```
bnoordhuis: Ditto.

src/module.cc
```diff
@@ -0,0 +1,388 @@
...
+ *
+ ******************************************************************************/
+NAN_METHOD(TriggerReport) {
+  Nan::HandleScope scope;
+  v8::Isolate* isolate = info.GetIsolate();
+  char filename[NR_MAXNAME + 1] = "";
+
+  if (info[0]->IsString()) {
+    // Filename parameter supplied
+    Nan::Utf8String filename_parameter(info[0]->ToString());
```
bnoordhuis: You don't have to ToString() the argument to Nan::Utf8String.

src/module.cc
```diff
@@ -0,0 +1,388 @@
...
+    if (filename_parameter.length() < NR_MAXNAME) {
+      snprintf(filename, sizeof(filename), "%s", *filename_parameter);
+    } else {
+      Nan::ThrowSyntaxError("nodereport: filename parameter is too long");
```
bnoordhuis: SyntaxError is kind of weird to throw for a bad argument.

src/module.cc
```diff
@@ -0,0 +1,388 @@
...
+  }
+}
+
+/*******************************************************************************
+ * External JavaScript APIs for nodereport configuration
+ *
+ ******************************************************************************/
+NAN_METHOD(SetEvents) {
+  Nan::Utf8String parameter(info[0]);
+  v8::Isolate* isolate = node_isolate;
```
bnoordhuis: Why don't you use info.GetIsolate() here?

src/module.cc
```diff
@@ -0,0 +1,388 @@
...
+    raise(SIGABRT); // core dump requested (default)
+  } else {
+    exit(1); // user specified that no core dump is wanted, just exit
+  }
+}
+
+bool OnUncaughtException(v8::Isolate* isolate) {
+   // Trigger NodeReport if required
+  if (nodereport_events & NR_EXCEPTION) {
+    TriggerNodeReport(isolate, kException, "exception", "OnUncaughtException (nodereport/src/module.cc)", NULL);
```
bnoordhuis: `__FILE__`?

src/module.cc
```diff
@@ -0,0 +1,388 @@
...
+  } 
+  if (nodereport_core) {
+    return true;
+  } else {
+    return false;
+  }
+}
```
bnoordhuis: Just `return nodereport_core;`?

src/module.cc
```diff
@@ -0,0 +1,388 @@
...
+
+#ifndef _WIN32
+static void SignalDumpInterruptCallback(Isolate *isolate, void *data) {
```
bnoordhuis: Somewhat schizophrenic star style.  It leans to the right here but to the left in most other places.

src/module.cc
```diff
@@ -0,0 +1,388 @@
...
+                        node::signo_string(report_signal),
+                        "node::SignalDumpInterruptCallback()", NULL);
+    }
+    report_signal = 0;
+  }
+}
+static void SignalDumpAsyncCallback(uv_async_t* handle) {
+  if (report_signal != 0) {
+    if (nodereport_verbose) {
+      fprintf(stdout,"nodereport: SignalDumpAsyncCallback handling signal\n");
```
bnoordhuis: Space after comma.

src/module.cc
```diff
@@ -0,0 +1,388 @@
...
+    }
+    if (nodereport_events & NR_SIGNAL) {
+      if (nodereport_verbose) {
+        fprintf(stdout,"nodereport: SignalDumpAsyncCallback triggering NodeReport\n");
```
bnoordhuis: Ditto.

src/module.cc
```diff
@@ -0,0 +1,388 @@
...
+  memset(&sa, 0, sizeof(sa));
+  sa.sa_handler = handler;
+  sigfillset(&sa.sa_mask);  // mask all signals while in the handler
+  sigaction(signal, &sa, &saved_sa);
+}
+
+// Utility function to restore an OS signal handler to its previous setting
+static void RestoreSignalHandler(int signal) {
+  sigaction(signal, &saved_sa, nullptr);
+}
```
bnoordhuis: It's arguably better API design to have the caller pass in `saved_sa` explicitly in RegisterSignalHandler() and RestoreSignalHandler().  It's easy to get fooled into thinking that it restores the handler for signal `signal` but that's not exactly what it does.

bnoordhuis: As an aside: maybe `s/signal/signo/` because that a) doesn't shadow the libc function of the same name, and b) is consistent with the function below.

src/module.cc
```diff
@@ -0,0 +1,388 @@
...
+    fprintf(stdout, "nodereport: initialization complete, event flags: %#x core flag: %#x signal flag: %#x\n",
+            nodereport_events, nodereport_core, nodereport_signal);
+#endif
+  }
+}
+
+NODE_MODULE(nodereport, Initialize)
+
+}  // namespace nodereport
+
```
bnoordhuis: General observation: there is a somewhat inconsistent mix of nullptr vs. NULL in this file.  Since you're going for C++11, I'd use nullptr everywhere.

src/node_report.cc
```diff
@@ -0,0 +1,808 @@
...
+static void PrintJavaScriptStack(FILE *fp, Isolate* isolate, DumpEvent event, const char *location);
+static void PrintStackFromStackTrace(FILE* fp, Isolate* isolate, DumpEvent event);
+static void PrintStackFrame(FILE* fp, Isolate* isolate, Local<StackFrame> frame, int index, void *pc);
+static void PrintNativeStack(FILE *fp);
+#ifndef _WIN32
+static void PrintResourceUsage(FILE *fp);
+#endif
+static void PrintGCStatistics(FILE *fp, Isolate* isolate);
+static void PrintSystemInformation(FILE *fp, Isolate* isolate);
+static void WriteInteger(FILE *fp, size_t value);
```
bnoordhuis: The star-leans-left/star-leans-right dichotomy exists in this file as well.

src/node_report.cc
```diff
@@ -0,0 +1,808 @@
...
+    } else {
+      fprintf(stderr, "Unrecognised argument for nodereport events option: %s\n", cursor);
+       return 0;
+    }
+    if (*cursor == '+') {
+      cursor++;  // Hop over the '+' separator
+    }
+  }
+  return event_flags;
+}
```
bnoordhuis: Indentation seems off in this function.

src/node_report.cc
```diff
@@ -0,0 +1,808 @@
...
+    // Construct the NodeReport filename, with timestamp, pid and sequence number
+    snprintf(filename, sizeof(filename), "%s", "NodeReport");
+    seq++;
+
+#ifdef _WIN32
+    snprintf(&filename[strlen(filename)], sizeof(filename) - strlen(filename),
+             ".%4d%02d%02d.%02d%02d%02d.%d.%03d.txt",
+             tm_struct.wYear, tm_struct.wMonth, tm_struct.wDay,
+             tm_struct.wHour, tm_struct.wMinute, tm_struct.wSecond,
+            pid, seq);
```
bnoordhuis: Indentation.

src/node_report.cc
```diff
@@ -0,0 +1,808 @@
...
+  } else {
+    // Regular file. Append filename to directory path if one was specified
+    if (strlen(report_directory) > 0) {
+      char pathname[NR_MAXPATH + NR_MAXNAME + 1] = "";
+      snprintf(pathname, sizeof(pathname), "%s", report_directory);
+#ifdef _WIN32
+      strcat(pathname, "\\");
+#else
+      strcat(pathname, "/");
+#endif
```
bnoordhuis: Potential buffer overrun here.

src/node_report.cc
```diff
@@ -0,0 +1,808 @@
...
+  // Print operating system information
+  PrintSystemInformation(fp, isolate);
+
+  fprintf(fp, "\n================================================================================\n");
+  fflush(fp);
+  fclose(fp);
+
+  fprintf(stderr, "Node.js report completed\n");
+  if (name != NULL) {
+    snprintf(name, NR_MAXNAME + 1, "%s", filename);  // return the NodeReport file name
```
bnoordhuis: Declare `name` like this in the function's parameter list: `char (*name)[NR_MAXNAME + 1]`.  That way the caller cannot pass in a wrong-sized buffer.

src/node_report.cc
```diff
@@ -0,0 +1,808 @@
...
+  fprintf(fp, "\nNode.js version: %s\n", NODE_VERSION);
+  fprintf(fp, "(v8: %s, libuv: %s, zlib: %s, ares: %s)\n",
+          V8::GetVersion(), uv_version_string(), ZLIB_VERSION, ARES_VERSION_STR);
+
+  // Print operating system and machine information (Windows)
+#ifdef _WIN32
+  fprintf(fp,"\nOS version: Windows ");
+#if defined(_MSC_VER) && (_MSC_VER >= 1900)
+  if (IsWindows1OrGreater()) {
+    fprintf(fp,"10 ");
```
bnoordhuis: Space after comma, here and elsewhere.

src/node_report.cc
```diff
@@ -0,0 +1,808 @@
...
+#else  // Unix, OSX
+  switch (event) {
+  case kException:
+  case kJavaScript:
+    // Print the stack using Message::PrintCurrentStackTrace() API
+    Message::PrintCurrentStackTrace(isolate, fp);
+    break;
+  case kFatalError:
+#if NODE_VERSION_AT_LEAST(6, 0, 0)
+    if (!strncmp(location,"MarkCompactCollector", sizeof("MarkCompactCollector") - 1)) {
```
bnoordhuis: Ditto.

src/node_report.cc
```diff
@@ -0,0 +1,808 @@
...
+  HANDLE hProcess = GetCurrentProcess();
+  SymSetOptions(SYMOPT_LOAD_LINES | SYMOPT_UNDNAME | SYMOPT_DEFERRED_LOADS);
+  SymInitialize(hProcess, NULL, TRUE);
+
+  WORD numberOfFrames = CaptureStackBackTrace(2, 64, frames, NULL);
+  
+  // Walk the frames printing symbolic information if available
+  for (int i = 0; i < numberOfFrames; i++) {
+    DWORD64 dwOffset64 = 0;
+    DWORD64 dwAddress = (DWORD64)(frames[i]);
```
bnoordhuis: C++-style casts, please!

src/node_report.cc
```diff
@@ -0,0 +1,808 @@
...
+    char buffer[sizeof(SYMBOL_INFO) + MAX_SYM_NAME * sizeof(TCHAR)];
+    PSYMBOL_INFO pSymbol = (PSYMBOL_INFO)buffer;
+    pSymbol->SizeOfStruct = sizeof(SYMBOL_INFO);
+    pSymbol->MaxNameLen = MAX_SYM_NAME;
+
+    if (SymFromAddr(hProcess, dwAddress, &dwOffset64, pSymbol)) {
+        DWORD dwOffset = 0;
+        IMAGEHLP_LINE64 line;
+        line.SizeOfStruct = sizeof(IMAGEHLP_LINE64);
```
bnoordhuis: `sizeof(line)`?

src/node_report.cc
```diff
@@ -0,0 +1,808 @@
...
+#ifdef _WIN32
+  fprintf(fp, "\nEnvironment variables\n");
+  LPTSTR lpszVariable;
+  LPTCH lpvEnv;
+
+  // Get pointer to the environment block
+  lpvEnv = GetEnvironmentStrings();
+  if (lpvEnv != NULL) {
+    // Variable strings are separated by NULL byte, and the block is terminated by a NULL byte.
+    lpszVariable = (LPTSTR) lpvEnv;
```
bnoordhuis: Ditto.

src/node_report.h
```diff
@@ -0,0 +1,75 @@
...
+void SetLoadTime();
+
+// secure_getenv() only available in glibc, revert to getenv() otherwise
+#if defined(__GLIBC__)
+#if !__GLIBC_PREREQ(2, 17)
+#define secure_getenv getenv
+#endif  // !__GLIBC_PREREQ(2, 17)
+#else
+#define secure_getenv getenv
+#endif  // defined(__GLIBC__)
```
bnoordhuis: Potential downside is that a .node compiled against glibc 2.17 won't work with e.g. glibc 2.15.  That is, you won't be able to simply copy the shared object around.

src/node_report.cc
rnchamberlain: https://msdn.microsoft.com/en-us/library/windows/desktop/ms680190%28v=vs.85%29.aspx

src/node_report.h
rnchamberlain: I wonder if that's a common thing to do, copy binaries around from an npm install. Someone is bound to do it I suppose, and they will get an unresolved symbol at load time. Under consideration! 

I think I have fixed everything else, see commits below, and thanks for the review.

src/node_report.cc
bnoordhuis: Right, but that doesn't preclude `sizeof(line)`, does it?

src/node_report.h
sam-github: Its a nice to have to be able to build on one system, and run on another slightly different one. It fails often enough that its usually necessary to build on a system of identical os/etc level and architecture as the deployment/production system.

src/node_report.cc
rnchamberlain: Either sizeof(<type>) or sizeof(<variable>) work OK I believe, the MSDN examples uses sizeof(<type>). Is sizeof(<variable>) preferred?

rnchamberlain:
Looking back we have an LGTM and an API LGTM on this, I'll merge shortly unless anyone objects. Thanks.

PR #5, opened 2016-10-27
Update documentation (README and package.json)

rnchamberlain:
Add version and platform support statements, make setEvents()
API more obvious, provide REPL example. For package.json correct
main, add bugs link, improve homepage link.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,9 +1,9 @@
 {
   "name": "nodereport",
-  "main": "build/Release/nodereport.node",
-  "version": "0.1.0",
+  "main": "nodereport.node",
+  "version": "1.0.3",
   "description": "Diagnostic NodeReport",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -14,8 +14,12 @@
   "dependencies": {
     "nan": "^2.3.5"
   },
+  "author": "Richard Chamberlain",
```
cjihrig: Since this is now a foundation project, maybe this should be `contributors` like `nan` does.

rnchamberlain: Ah yes, array of` "contributors": [ ` - thanks, will change

sam-github:
Would be better to separate into two commits, and have them describe the changes (like `package: set metadata (author, homepage, version, ...)`)

rnchamberlain:
OK, I will raise a new PR with two separate commits.

PR #8, opened 2016-10-27
Add ignore files

sam-github:


.npmignore
```diff
@@ -0,0 +1,8 @@
+NodeReport*.txt
```
cjihrig: What if you blacklist everything by default and then add things back. Something like this could work:

```
*
!src/**
!binding.gyp
```

cjihrig: Yes, npm always includes certain files (sorry, I can't get any more specific of a link than [this](https://docs.npmjs.com/misc/developers#keeping-files-out-of-your-package)). I'm only throwing it out there as a suggestion because it is slightly more future proof. I don't have a problem with the approach here.

sam-github: I tried it, and liked it.

rnchamberlain:
Good stuff, could we have a comment in .npmignore just to explain that README/LICENSE etc get published anyway, as that's not obvious. Also, note I added the `"test": "node test/autorun.js"` entry into package.son on request, so that people could run "npm test", see https://docs.npmjs.com/cli/test. So I think that the /test directory ought to be included in the npm publish?

rnchamberlain:
LGTM

Issue #4, opened 2016-10-26
Compilation error on x64 Linux with Node v6.9.1, gcc 4.8.5


richardlau:
```
-bash-4.2$ npm install

> nodereport@0.1.0 install /home/users/riclau/sandbox/github/nodereport
> node-gyp rebuild

make: Entering directory `/home/users/riclau/sandbox/github/nodereport/build'
make: Warning: File `nodereport.target.mk' has modification time 914 s in the future
  CXX(target) Release/obj.target/nodereport/src/node_report.o
../src/node_report.cc: In function âvoid nodereport::PrintSystemInformation(FILE*, v8::Isolate*)â:
../src/node_report.cc:765:27: error: expected â)â before âPRIu64â
         fprintf(fp, "%16" PRIu64, limit.rlim_cur);
                           ^
../src/node_report.cc:765:49: warning: conversion lacks type at end of format [-Wformat=]
         fprintf(fp, "%16" PRIu64, limit.rlim_cur);
                                                 ^
../src/node_report.cc:765:49: warning: too many arguments for format [-Wformat-extra-args]
../src/node_report.cc:770:27: error: expected â)â before âPRIu64â
         fprintf(fp, "%16" PRIu64 "\n", limit.rlim_max);
                           ^
../src/node_report.cc:770:54: warning: conversion lacks type at end of format [-Wformat=]
         fprintf(fp, "%16" PRIu64 "\n", limit.rlim_max);
                                                      ^
../src/node_report.cc:770:54: warning: too many arguments for format [-Wformat-extra-args]
make: *** [Release/obj.target/nodereport/src/node_report.o] Error 1
make: Leaving directory `/home/users/riclau/sandbox/github/nodereport/build'
gyp ERR! build error
gyp ERR! stack Error: `make` failed with exit code: 2
gyp ERR! stack     at ChildProcess.onExit (/dev/shm/usenode.riclau/node-v6.9.1-linux-x64/lib/node_modules/npm/node_modules/node-gyp/lib/build.js:276:23)
gyp ERR! stack     at emitTwo (events.js:106:13)
gyp ERR! stack     at ChildProcess.emit (events.js:191:7)
gyp ERR! stack     at Process.ChildProcess._handle.onexit (internal/child_process.js:215:12)
gyp ERR! System Linux 3.10.0-327.18.2.el7.x86_64
[...]

PR #9, opened 2016-10-28
Need to define __STDC_FORMAT_MACROS for some glibc versions

rnchamberlain:
Fix for https://github.com/nodejs/nodereport/issues/4. The compilation failure was caused by undefined macro  `PRIu64`. This macro is defined in the include file inttypes.h, but some levels of glibc required that `#define __STDC_FORMAT_MACROS` is also set, eg see http://en.cppreference.com/w/cpp/types/integer

### DIFF
--- a/src/node_report.cc
+++ b/src/node_report.cc
@@ -22,6 +22,10 @@
 #else
 #include <sys/time.h>
 #include <sys/resource.h>
+// Get the standard printf format macros for C99 stdint types.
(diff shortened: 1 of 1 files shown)


sam-github:
LGTM

sam-github:
@nodejs/post-mortem A post-mortem collaborator is required to give a LGTM for this PR, could one of you look at it?

mhdawson:
LGTM

PR #11, opened 2016-10-28
Fix nodereport requires

### LINKED ISSUE #7: tests all fail
```
core/nodereport (master *% u=) % npm test

> nodereport@0.1.0 test /home/sam/w/core/nodereport
> node test/autorun.js

autorun.js: running child process #1 to produce NodeReport on API call
autorun.js: running child process #2 to produce NodeReport on exception
autorun.js: running child process #3 to produce NodeReport on fatal error
autorun.js: running child process #4 to produce NodeReport on SIGUSR2
autorun.js: test results: 
        test 1 :  fail
        test 2 :  fail
        test 3 :  fail
        test 4 :  fail

assert.js:85
  throw new assert.AssertionError({
  ^
AssertionError: false == true
    at ChildProcess.<anonymous> (/home/sam/w/core/nodereport/test/autorun.js:27:10)
    at emitTwo (events.js:106:13)
    at ChildProcess.emit (events.js:191:7)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:215:12)

```

sam-github:
The tests would only run when nodereport was itself git cloned into a directory called `node_modules`, tricking the require system into thinking nodereport was a dependency of itself.

Partially fixes #7

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,3 +41,4 @@
 NodeReport*.txt
 npm-debug.log
 nodereport-*.tgz
+nodereport_test.log
--- a/test/test_1.js
+++ b/test/test_1.js
@@ -1,5 +1,5 @@
 // NodeReport API example
-var nodereport = require('nodereport');
+var nodereport = require('../');
 
 console.log('api_call.js: triggering a NodeReport via API call...');
 
--- a/test/test_2.js
+++ b/test/test_2.js
@@ -1,5 +1,5 @@
 // Testcase to produce an uncaught exception
-require('nodereport').setEvents("exception");
+require('../').setEvents("exception");
 
 console.log('exception.js: throwing an uncaught user exception....');
 
--- a/test/test_3.js
+++ b/test/test_3.js
@@ -1,5 +1,5 @@
 // Testcase to produce a fatal error (javascript heap OOM)
-require('nodereport').setEvents("fatalerror");
+require('../').setEvents("fatalerror");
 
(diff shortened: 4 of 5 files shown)


rnchamberlain:
I tried it and it works for me, LGTM

sam-github:
@rnchamberlain as one of the core collaborators, you can press the merge button as soon as you have LGTMed, and its sat for 48 hours for someone else to say 👎  if they want. I don't have merge auth.

mhdawson:
LGTM

rnchamberlain:
@sam-github one more question on this: how do we run the tests against an 'npm installed' nodereport module? I think that's a reasonable scenario. So we get the tests now by cloning the repo from github, but the relative path require() only works if we 'node-gyp build' that cloned nodereport. One way to test the original npm install is to copy -R the tests folder from the git clone across to node_modules/nodereport, but that seems clunky to me. Thoughts?

sam-github:
> I think that's a reasonable scenario.

Why? What are you testing? npm is for delivering requireable dependencies, not for downloading project source for development. This seems to me to be a test that when you npm publish/install, the output files are the same as the input files. Its a test of npm-the-tool, not a test of the package.

When you get a python module, does it come with tests? What about a java jar file or compiled C library? They don't come with tests either, the tests are run on the built artifacts before pushing them into whatever packaging system is used to distribute the artifacts.

Or take node, I don't think you can run its tests against the node installed in the system, though maybe there is some hack I'm unaware of.

PR #3, opened 2016-10-25
Update README and package.json for NPM publish

rnchamberlain:
Improve README to indicate supported OS and Node.js versions, provide REPL example and clarify usage. Update version and add github links to package.json.

PR #14, opened 2016-11-01
Prototype fix for failures in nodereport test #3 - do not land

rnchamberlain:
The v8 PrintCurrentStackTrace() API cannot be used to collect the JS stack if we are actually inside a v8 GC when the fatal error (and NodeReport) is triggered. This happens intermittently for OOM failures, and the API will crash or hang. The explanation is that objects are on the move, see https://groups.google.com/forum/#!topic/v8-dev/c6vGpNMScKU. The problem also shows up in the stderr failure messages from node, the usual JS stack trace is replaced by a "Cannot get stack trace in GC" message.

The workaround in nodereport is to detect that we are in GC and skip the stack trace. This PR adds another case to to the check we use to detect we are in GC.

### DIFF
--- a/src/node_report.cc
+++ b/src/node_report.cc
@@ -436,7 +436,8 @@ static void PrintJavaScriptStack(FILE* fp, Isolate* isolate, DumpEvent event, co
     break;
   case kFatalError:
 #if NODE_VERSION_AT_LEAST(6, 0, 0)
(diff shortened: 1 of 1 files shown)


rnchamberlain:
Investigating issues with JS stacktrace on fatal error in recent v6 versions. Closing this for now.

Issue #12, opened 2016-10-28
test 3 fails consistently


richardlau:
Node v6.9.1, x64 Linux RHEL 7.2.

Had to patch the nodereport source with nodejs/nodereport#9 to be able to compile. I also needed to checkout nodereport from github as the tests are no longer pulled down when I `npm install` (I think as a result of nodejs/nodereport#8).

```
-bash-4.2$ npm test -- --save_log

> nodereport@1.0.4 test /home/users/riclau/sandbox/scratch/test/pkgtest/node_modules/nodereport
> node test/autorun.js "--save_log"

autorun.js: running child process #1 to produce NodeReport on API call
autorun.js: running child process #2 to produce NodeReport on exception
autorun.js: running child process #3 to produce NodeReport on fatal error
autorun.js: running child process #4 to produce NodeReport on SIGUSR2
autorun.js: located NodeReport:  NodeReport.20161028.143703.61091.001.txt
autorun.js: located NodeReport:  NodeReport.20161028.143703.61097.001.txt
autorun.js: located NodeReport:  NodeReport.20161028.143704.61099.001.txt
autorun.js: located NodeReport:  NodeReport.20161028.143704.61104.001.txt
autorun.js: test results:
        test 1 :  pass
        test 2 :  pass
        test 3 :  fail
        test 4 :  pass
-bash-4.2$
```

[nodereport_test.log](https://gist.github.com/richardlau/28214a917a1db97f5ef2a1f65e01c69d)

richardlau:
I modified `test/autorun.js` so that it doesn't delete the report for test 3.
```
-bash-4.2$ npm test -- --save_log

> nodereport@1.0.4 test /home/users/riclau/sandbox/scratch/test/pkgtest/node_modules/nodereport
> node test/autorun.js "--save_log"

autorun.js: running child process #1 to produce NodeReport on API call
autorun.js: running child process #2 to produce NodeReport on exception
autorun.js: running child process #3 to produce NodeReport on fatal error
autorun.js: running child process #4 to produce NodeReport on SIGUSR2
autorun.js: located NodeReport:  NodeReport.20161028.145340.75220.001.txt
autorun.js: located NodeReport:  NodeReport.20161028.145341.75227.001.txt
autorun.js: located NodeReport:  NodeReport.20161028.145340.75226.001.txt
autorun.js: located NodeReport:  NodeReport.20161028.145341.75233.001.txt
autorun.js: test results:
        test 1 :  pass
[...]

PR #16, opened 2016-11-03
Fix for failure in fatal error (OOM) test

rnchamberlain:
Fix for https://github.com/nodejs/nodereport/issues/12

On v6.9.1 we get a second OOM in v8 if we call the v8 stacktrace API in the fatal error (OOM) scenario. It worked OK on v6.3.0. Fix is to remove the call for now, for the fatal error (OOM) scenario. There was an additional check to avoid the call if we failed actually inside GC, this is also removed as it is now redundant.

This means we won't provide a JS stacktrace in the NodeReport for this particular scenario.  The v8 stderr output does show a single JS frame, so we may be able to obtain that for nodereport via another route at some point. Alternatively we can investigate why the v8 API is failing and see if that can be improved.

### DIFF
--- a/src/node_report.cc
+++ b/src/node_report.cc
@@ -439,15 +439,7 @@ static void PrintJavaScriptStack(FILE* fp, Isolate* isolate, DumpEvent event, co
     Message::PrintCurrentStackTrace(isolate, fp);
     break;
   case kFatalError:
-#if NODE_VERSION_AT_LEAST(6, 0, 0)
-    if (!strncmp(location, "MarkCompactCollector", sizeof("MarkCompactCollector") - 1)) {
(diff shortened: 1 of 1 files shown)


mhdawson:
While I agree that investigating to see if we can provide a stack trace, this change seems reasonable so that we don't have test failures and then we can add back what is necessary when/if we can provide the stack trace.

mhdawson:
Unless there are objections before then will plan to land tomorrow.

richardlau:
No objections here. Be good to get the tests passing :smile:

mhdawson:
Landed as 94a1125f92bc50adcf046f06168a5724c01b72f6

PR #17, opened 2016-11-04
Add support for nodereport on AIX

rnchamberlain:
1. features.h and execinfo.h do not exist on AIX
2. backtrace() API is not available, native stacks disabled in nodereport for now
3. ru_utime.tv_usec is int not long (same as OSX)
4. ulimit RLIMIT_MEMLOCK is not available on AIX
5. PRIu64 macro is not available in gcc on AIX

@richardlau FYI

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ It includes JavaScript and native stack traces, heap statistics,
 platform information and resource usage etc. With the report enabled,
 reports can be triggered on unhandled exceptions, fatal errors, signals
 and calls to a JavaScript API. The module supports Node.js v4, v6 and v7
-on Linux, MacOS and Windows.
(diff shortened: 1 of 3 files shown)


mhdawson:
Unless there are any objections will plan to land this tomorrow.

richardlau:
Changes look good. Compilation is successful and tests are passing with the exception of test 3, which I believe will be addressed by #16.

richardlau:
Maybe also update `README.md` to list AIX as a supported platform?

rnchamberlain:
Rebased and updated README to list AIX as supported.

mhdawson:
Landed as bcf3a336208ef8e2bd9fc975d4931e0e5e65409b

PR #20, opened 2016-11-08
Improvement to Windows version reporting

richardlau:
Changes detection of Windows version from using the helper functions in [VersionHelpers.h](https://msdn.microsoft.com/en-us/library/windows/desktop/dn424972%28v=vs.85%29.aspx) to use [NetServerGetInfo](https://msdn.microsoft.com/en-us/library/windows/desktop/aa370624%28v=vs.85%29.aspx) instead.

[IsWindows10OrGreater](https://msdn.microsoft.com/en-us/library/windows/desktop/dn905474%28v=vs.85%29.aspx) requires Windows SDK 10 (I was getting compilation errors because it was a) spelt incorrectly and b) I didn't have Windows SDK 10 installed even though I was compiling with Visual Studio 2015). It also doesn't always return true even if the system is Windows 10:
`Applications not manifested for Windows 10 return false, even if the current operating system version is Windows 10.`

Mappings of major and minor versions reported by `NetServerGetInfo` to more common Windows version names are based on ~~[env plugin from omr-agentcore](https://github.com/RuntimeTools/omr-agentcore/blob/3.0.10/src/ibmras/monitoring/plugins/common/environment/envplugin.cpp#L493)~~ [this table on MSDN](https://msdn.microsoft.com/en-gb/library/windows/desktop/ms724832(v=vs.85).aspx) -- I've tested on Windows 7, Windows 10 and Server 2016 and the generated report contains the expected values. 

Operating system|Version number
---|---
Windows 10|10.0
Windows Server 2016|10.0
Windows 8.1|6.3
Windows Server 2012 R2|6.3
Windows 8|6.2
Windows Server 2012|6.2
Windows 7|6.1
Windows Server 2008 R2|6.1
Windows Server 2008|6.0
Windows Vista|6.0
Windows Server 2003 R2|5.2
Windows Server 2003|5.2
Windows XP 64-Bit Edition|5.2
Windows XP|5.1
Windows 2000|5.0

@rnchamberlain

### DIFF
--- a/src/node_report.cc
+++ b/src/node_report.cc
@@ -18,7 +18,7 @@
 #include <Windows.h>
 #include <process.h>
 #include <dbghelp.h>
-#include <VersionHelpers.h>
+#include <Lm.h>
 #else
 #include <sys/time.h>
 #include <sys/resource.h>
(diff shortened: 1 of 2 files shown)


src/node_report.cc
```diff
@@ -376,28 +376,53 @@ static void PrintVersionInformation(FILE* fp) {
...
+                }
+                break;
+        case 6: switch (pBuf->sv101_version_minor) {
+                  case 0: fprintf(fp, "%s\n", (isServer ? "Server 2008"
+                                                        : "Vista"));
+                          break;
+                  case 1: fprintf(fp, "%s\n", (isServer ? "Server 2008 R2"
+                                                        : "7"));
+                          break;
+                  case 2: fprintf(fp, "%s\n", (isServer ? "Server 8"
```
richardlau: This should be `Server 2012`. Will change tomorrow.

richardlau: Also add `case 3` for `Server 2012 R2` and `8.1`.

richardlau: ref: https://msdn.microsoft.com/en-gb/library/windows/desktop/ms724832(v=vs.85).aspx

richardlau:
[Windows operating system version to name mapping on MSDN](https://msdn.microsoft.com/en-gb/library/windows/desktop/ms724832(v=vs.85).aspx)

rnchamberlain:
LGTM - tested OK on Windows 7

Minor:
need spaces round the '&' here
   `major = pBuf->sv101_version_major&MAJOR_VERSION_MASK;`

indentation on the multi-level switch statements is a bit wide, personally i'd prefer
```
switch (major) {
case 5: 
  switch (pBuf->sv101_version_minor) {
  case 0:
    fprintf(....
```

richardlau:
@rnchamberlain Fixed up the minor nit and redid the switch statement so that it's easier to read now. Tested on Windows 7 and Windows Server 2012 R2 this time.

richardlau:
Rebased onto current master.

rnchamberlain:
Landed as https://github.com/nodejs/nodereport/commit/ef67b730ffc7e94e1918b940ac406a52a7a36a14

Issue #6, opened 2016-10-27
illegal hardware exception


sam-github:
Is this expected?

```
core/nodereport (master % u=) % node demo/exception.js                  
exception.js: Node running
exception.js: Go to http://<machine>:8080/ or http://localhost:8080/

Writing Node.js report to file: NodeReport.20161027.100601.28911.001.txt
Node.js report completed
Uncaught #<UserException>

FROM
Server.my_listener (/home/sam/w/core/nodereport/demo/exception.js:15:5)
emitTwo (events.js:106:13)
Server.emit (events.js:191:7)
HTTPParser.parserOnIncoming [as onIncoming] (_http_server.js:546:12)
HTTPParser.parserOnHeadersComplete (_http_common.js:99:23)
zsh: illegal hardware instruction (core dumped)  node demo/exception.js
```

System info:
- linux ubuntu x64
- node 6.9.1, npm 3.10.8
- nodereport from git 7cd3f76f7114ec369aec965c133626293f246683

I got this when I was trying to verify whether nodereport after reporting an uncaught exception, would let the exception continue to propagate and cause node to exit (as it should).

sam-github:
Its fairly reproduceable. I ran with gdb, too:

```
Node.js report completed
Uncaught #<UserException>

FROM
Server.my_listener (/home/sam/w/core/nodereport/demo/exception.js:15:5)
emitTwo (events.js:106:13)
Server.emit (events.js:191:7)
HTTPParser.parserOnIncoming [as onIncoming] (_http_server.js:546:12)
HTTPParser.parserOnHeadersComplete (_http_common.js:99:23)

Thread 1 "node" received signal SIGILL, Illegal instruction.
0x000000000132de89 in v8::base::OS::Abort() ()
(gdb) bt
#0  0x000000000132de89 in v8::base::OS::Abort() ()
#1  0x0000000000d4e30a in v8::internal::Isolate::Throw(v8::internal::Object*, v8::internal::MessageLocation*) ()
#2  0x0000000000eaf423 in v8::internal::Runtime_Throw(int, v8::internal::Object**, v8::internal::Isolate*) ()
```

rnchamberlain:
[...]

PR #24, opened 2016-11-17
Fix test-exception.js for PPC

richardlau:
On PPC architectures, `--abort_on_uncaught_exception` (currently set by nodereport for the `exception` event, see #6) exits with SIGTRAP instead of SIGILL (nodejs/node#3239). This PR fixes the `test-exception.js` testcase for the current behaviour.

### DIFF
--- a/test/common.js
+++ b/test/common.js
@@ -17,6 +17,14 @@ exports.findReports = (pid) => {
   return files.filter((file) => filePattern.test(file));
 };
 
+exports.isPPC = () => {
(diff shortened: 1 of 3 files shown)


rnchamberlain:
LGTM

rnchamberlain:
Landed as https://github.com/nodejs/nodereport/commit/60653831d94609a41ca1e2fcb4a02b7b5ed50b3b

@richardlau fyi there was a trailing whitespace fix in test-exception.js

Issue #26, opened 2016-11-17
triggerReport() fails on x86 Windows


richardlau:
Node v6.9.1 win-x86

Reproduced on Windows 7 (Visual Studio 2012 or 2015) and Windows 10 (Visual Studio 2015)

Looks like the report is truncated, exit code suggests an ACCESS_VIOLATION.
```
C:\work\node\github\nodereport>npm test

> nodereport@1.0.5 test C:\work\node\github\nodereport
> tap test/test*.js

test/test-api.js ...................................... 4/7
  not ok Process exited cleanly
    +++ found
    --- wanted
    -0
    +3221225477
    compare: ===
    at:
      line: 15
      column: 9
      file: test/test-api.js
      type: ChildProcess
      function: child.on
    stack: |
      ChildProcess.child.on (test/test-api.js:15:9)
    source: |
      tap.equal(code, 0, 'Process exited cleanly');

  Validating NodeReport.20161117.185734.4656.001.txt
  not ok Checking report contains JavaScript Heap section
    found: "====================================================================
============\r\n==== NodeReport ================================================
================\r\n\r\nEvent: JavaScript API, location: \"nodereport::TriggerRe
port\"\r\nFilename: NodeReport.20161117.185734.4656.001.txt\r\nDump event time:
 2016/11/17 18:57:34\r\nModule load time: 2016/11/17 18:57:34\r\nProcess ID: 465
6\r\n\r\nNode.js version: v6.9.1\r\n(v8: 5.1.281.84, libuv: 1.9.1, zlib: 1.2.8,
ares: 1.10.1-DEV)\r\n\r\nOS version: Windows 7\r\n\r\nMachine: IBM826-R93F9G4 \r
\n\r\n==========================================================================
======\r\n==== JavaScript Stack Trace ==========================================
==========\r\n\r\nJavaScript VM state: EXTERNAL\r\n\r\n 0: [pc=0x1FB353CE] C:\\w
ork\\node\\github\\nodereport\\test\\test-api.js:6:14\r\n 1: [pc=0x1FB35128] Mod
ule._compile (module.js:570:32)\r\n 2: [pc=0x1FB2FC8E] Module._extensions..js (m
odule.js:579:10)\r\n 3: [pc=0x1FB2E19A] Module.load (module.js:487:32)\r\n 4: [p
c=0x1FB2DE1A] tryModuleLoad (module.js:446:12)\r\n 5: [pc=0x1FB255D9] Module._lo
[...]

PR #27, opened 2016-11-18
Fix fprintf calls on Windows

richardlau:
`pSymbol->Address` is `ULONG64` but `fprintf`'s `%p` format on x86 expects 32-bits.

Also wrapped the long edited lines.

Fixes nodejs/nodereport#26

### DIFF
--- a/src/node_report.cc
+++ b/src/node_report.cc
@@ -598,21 +598,20 @@ void PrintNativeStack(FILE* fp) {
     pSymbol->MaxNameLen = MAX_SYM_NAME;
 
     if (SymFromAddr(hProcess, dwAddress, &dwOffset64, pSymbol)) {
-        DWORD dwOffset = 0;
-        IMAGEHLP_LINE64 line;
-        line.SizeOfStruct = sizeof(line);
-        if (SymGetLineFromAddr64(hProcess, dwAddress, &dwOffset, &line)) {
-          fprintf(fp, "%2d: [pc=0x%p] %s [+%d] in %s: line: %lu\n", i, pSymbol->Address, pSymbol->Name, dwOffset, line.FileName, line.LineNumber);
-        } else {
-          // SymGetLineFromAddr64() failed, just print the address and symbol
-          if (dwOffset64 <= 32) { // sanity check
-            fprintf(fp, "%2d: [pc=0x%p] %s [+%d]\n", i, pSymbol->Address, pSymbol->Name, dwOffset64);
-          } else {
-            fprintf(fp, "%2d: [pc=0x%p]\n", i, pSymbol->Address);
-          }
-        }
+      DWORD dwOffset = 0;
+      IMAGEHLP_LINE64 line;
+      line.SizeOfStruct = sizeof(line);
+      if (SymGetLineFromAddr64(hProcess, dwAddress, &dwOffset, &line)) {
+        fprintf(fp, "%2d: [pc=0x%p] %s [+%d] in %s: line: %lu\n", i,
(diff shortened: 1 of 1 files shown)


src/node_report.cc
```diff
@@ -602,17 +602,23 @@ void PrintNativeStack(FILE* fp) {
...
           } else {
-            fprintf(fp, "%2d: [pc=0x%p]\n", i, pSymbol->Address);
+            fprintf(fp, "%2d: [pc=0x%p]\n", i, 
+			  reinterpret_cast<void*>(pSymbol->Address));
           }
         }
     } else { // SymFromAddr() failed, just print the address
-      fprintf(fp, "%2d: [pc=0x%p]\n", i, pSymbol->Address);
+      fprintf(fp, "%2d: [pc=0x%p]\n", i, 
+		reinterpret_cast<void*>(pSymbol->Address));
```
richardlau: If `SymFromAddr()` fails, is `pSymbol->Address` valid here?
@rnchamberlain

rnchamberlain: I suspect that line was copied from 4 lines above, where SymFromAddr() has succeeded. The SymFromAddr() call might copy the address into pSymbol->Address in failure cases, but yes, it would be better/more logical to use dwAddress instead at this point. Can you add that to your fix? Thanks.

richardlau: done

src/node_report.cc
```diff
@@ -602,17 +602,22 @@ void PrintNativeStack(FILE* fp) {
         IMAGEHLP_LINE64 line;
         line.SizeOfStruct = sizeof(line);
         if (SymGetLineFromAddr64(hProcess, dwAddress, &dwOffset, &line)) {
-          fprintf(fp, "%2d: [pc=0x%p] %s [+%d] in %s: line: %lu\n", i, pSymbol->Address, pSymbol->Name, dwOffset, line.FileName, line.LineNumber);
+          fprintf(fp, "%2d: [pc=0x%p] %s [+%d] in %s: line: %lu\n", i, 
+			reinterpret_cast<void*>(pSymbol->Address), pSymbol->Name, 
+			dwOffset, line.FileName, line.LineNumber);
         } else {
           // SymGetLineFromAddr64() failed, just print the address and symbol
           if (dwOffset64 <= 32) { // sanity check
```
richardlau: @rnchamberlain Is this sanity check valid? It seems to be consistently failing for me and we drop into the else clause.

richardlau: nb. Consistently dropping into the else clause here is why only `test-api.js` is failing in nodejs/nodereport#26 as it's succeeding the `SymGetLineFromAddr64` call -- The other tests fail the call and then fail the sanity check. The original `fprintf` for the else clause of the sanity check doesn't have parameters after the `pSymbol->Address` argument, so there the `fprintf` call there didn't crash.

richardlau:
Updated. I've dropped the sanity check as it seemed wrong and prevents the name of the method being printed.

rnchamberlain:
@richardlau I can't remember why that `if (dwOffset64 <= 32) ` sanity check was in there. It's checking the offset from the symbol (the displacement parameter returned by SymFromAddr()). These are function-name symbols we are getting, and it seems that the address could be some way down in the function. So checking for <= 32 is clearly too limiting. Agree to drop it, we'll keep a lookout for bogus symbols appearing. Planning to land this PR tomorrow.

rnchamberlain:
Landed as https://github.com/nodejs/nodereport/commit/d988b6cf963f18e999db8aacc723fea754191663

Issue #25, opened 2016-11-17
support `node -r nodereport app.js` as well as programmatic use via API


sam-github:
ATM, nodereport works better via API, its clumsy if you want automatic hooking, see: 

  https://github.com/nodejs/nodereport/pull/23#issuecomment-261267666

On the other hand, if you want to use nodereport to trigger a report manually, just using its public API, you don't want its (quite intrusive) hook mechanisms, which modify exit codes, behaviours, and even steal signals.

In summary, we need easy opt-in for default behaviour, but we also need easy opt-out, and I don't see any way to provide the two via a single require entry point that does *not* depend on env vars for configuration.

To make both use-cases easy, I suggest:

    node -r nodereport/attach app.js

for the opt-in case, and

    require("nodereport").triggerReport()

for the opt-out.

I'm OK with the require paths being switched, so `node -r nodereport` opts-in, and `require("nodereport/api")` or something of the like just requires the nodereport API without hooks.

Any other suggestions?

rnchamberlain:
+1, with preference for 'paths being switched' because the API user is more likely to study the docs.

@sam-github do you have any documentation references for the 'module/option' syntax?

Fishrock123:
I would prefer it to be as easy to use on an existing script out of the box as possible.

Maybe it's just me but I don't see the code hooks as being quite as valuable 9and the feel a bit more invasive). IIRC this is the sort of reason `-r` exists in the first place.

FWIW tools like [dprof](https://github.com/AndreasMadsen/dprof) have this similar behavior and I feel a previous precedent has already been set by similar modules that act upon requiring.

sam-github:
So you two would prefer

    node -r nodereport app.js

for the opt-in case, and

    require("nodereport/api").triggerReport()

for opt-out of hooks, with the understanding that this works to get hooks, and have the API:

    require("nodereport").triggerReport()

Above is fine by me.

richardlau:
I think I'm also in favour of
```
node -r nodereport app.js
```
being opt-in and 
```
require("nodereport/api").triggerReport()
```
for opt-out of hooks.

Since this would be a change in behaviour, would this be semver-major?

sam-github:
yep, semver-major​

mhdawson:
Given the consensus, I'm ok with node -r being opt in by default.

rnchamberlain:
PR https://github.com/nodejs/nodereport/pull/33 landed.

[...]

PR #28, opened 2016-11-23
Opt-in hooks and signal by default

rnchamberlain:
This changes the default action on require('nodereport') to opt-in to the exception and fatal-error hooks, and to enable the signal handler. So 'node -r nodereport application.js' will get all the nodereport functionality without requiring application code changes.

For users programming to the API who don't want the invasive hooks and signal handler, a alternative api-only entry point, require('nodereport/api') is provided. This does not enable the hooks and signal unless and until .setEvents() is called.

Addresses issue https://github.com/nodejs/nodereport/issues/25


main.js
```diff
@@ -0,0 +1,34 @@
+// Main module entry point for nodereport
+
+var api = require('./api');
```
richardlau: `const` instead of `var`?

main.js
```diff
@@ -0,0 +1,34 @@
...
+
+// Process NODEREPORT_EVENTS env var
+var options = process.env.NODEREPORT_EVENTS;
```
richardlau: `const` here as well.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "nodereport",
-  "main": "nodereport.node",
+  "main": "./main",
```
richardlau: `./` is unnecessary here.

test/test-exception.js
```diff
@@ -21,14 +21,9 @@ if (process.argv[2] === 'child') {
...
-    const expectedExitCode = common.isWindows() ? 0xC0000005 : null;
-    const expectedSignal = common.isWindows() ? null :
-                           common.isPPC() ? 'SIGTRAP' : 'SIGILL';
-    tap.plan(4);
-    tap.equal(code, expectedExitCode, 'Process should not exit cleanly');
-    tap.equal(signal, expectedSignal,
-              'Process should exit with expected signal ');
+  child.on('exit', (code) => {
+    tap.plan(3);
+    tap.equal(code, 1, 'Process should exit with expected return code');
```
richardlau: Just tried running your branch on Node.js v6.9.1 linux-x64 and got this:
```
-bash-4.2$ npm test

> nodereport@1.0.6 test /home/users/riclau/sandbox/github/nodereport
> tap test/test*.js

test/test-api.js ...................................... 7/7
test/test-exception.js ................................ 6/7
  not ok Process should exit with expected return code
    +++ found
    --- wanted
    -1
    +0
    compare: ===
    at:
      line: 26
      column: 9
      file: test/test-exception.js
      type: ChildProcess
      function: child.on
    stack: |
      ChildProcess.child.on (test/test-exception.js:26:9)
    source: |
      tap.equal(code, 1, 'Process should exit with expected return code');

test/test-fatal-error.js .............................. 7/7
test/test-signal.js ................................... 8/8
total ............................................... 28/29


  28 passing (2s)
  1 failing

-bash-4.2$
```

rnchamberlain:
Ah, behavior on uncaught exception (when hooked) on v6.9.1 is different to v7.0.0. Return code is 0 instead of 1 - hence the test failure - and the exception message and stack trace on the console is missing. Investigating.

Re the tests, currently we are testing for example:` require('nodereport').setEvents('fatalerror'); `which is no longer as important a use case as 
1)` require('nodereport')` or 
2) `require('nodereport/api').setEvents('fatalerror');` 
We should probably test both of those cases, at least.

Yes docs will need updating.

Fishrock123:
I'll try to get to this soon. :D

.npmignore
```diff
@@ -1,3 +1,4 @@
 *
 !src/**
 !binding.gyp
+!main.js
```
sam-github: index.js is traditional, and doesn't require config in package.json

main.js
```diff
@@ -0,0 +1,34 @@
...
+if (options) {
+  api.setEvents(options);
+} else {
+  // Default action - all events enabled
+  api.setEvents('exception+fatalerror+signal+apicall');
+}
+
+exports.triggerReport = function() {
```
sam-github: why not `exports.triggerReport = api.triggerReport;`?

main.js
```diff
@@ -0,0 +1,34 @@
...
+};
+exports.setFileName = function() {
+  return api.setFileName.apply(null, arguments);
+};
+exports.setDirectory = function() {
+  return api.setDirectory.apply(null, arguments);
+};
+exports.setVerbose = function() {
+  return api.setVerbose.apply(null, arguments);
+};
```
sam-github: ^ no newline at end of file

main.js
rnchamberlain: done

main.js
rnchamberlain: done

package.json
rnchamberlain: Line now removed (on Sam's suggestion to use index.js instead of main.js)

test/test-exception.js
rnchamberlain: I have reverted the change to the behaviour on uncaught exception, I think that needs a separate PR. Test now passing for me.

.npmignore
rnchamberlain: done, using index.js now

main.js
rnchamberlain: done

rnchamberlain:
I have reverted the change to the default behavior on uncaught exception - currently it goes to the v8 abort/sigill - will fix later in a separate PR.

Docs updated to advertise `node -r nodereport app.js`

For the tests, I have changed them to test the new default behavior - i.e. setEvents() calls no longer needed. I will add tests specifically for `require('nodereport/api')` in a later PR.

Fishrock123:
Sounds good to me. Doesn't merge tho so I didn't try it.

rnchamberlain:
As there's been some churn on my branch I'll close this and raise a new PR.

Issue #29, opened 2016-11-24
Reported Node.js version is incorrect


richardlau:
Reports currently contain in the `NodeReport` section:
```
Node.js version: v6.9.1
(v8: 5.1.281.84, libuv: 1.9.1, zlib: 1.2.8, ares: 1.10.1-DEV)
```
If I compile nodereport with one version of Node.js (e.g. v6.9.1) and then run on another version (e.g. v6.7.0) the report incorrectly reports `v6.9.1` and not `v6.7.0` (note that it correctly reports `v8: 5.1.281.83`):
```
-bash-4.2$ node
> process.versions
{ http_parser: '2.7.0',
  node: '6.7.0',
  v8: '5.1.281.83',
  uv: '1.9.1',
  zlib: '1.2.8',
  ares: '1.10.1-DEV',
  icu: '57.1',
  modules: '48',
  openssl: '1.0.2j' }
> require('.').triggerReport()

Writing Node.js report to file: NodeReport.20161124.151013.8560.001.txt
Node.js report completed
'NodeReport.20161124.151013.8560.001.txt'
> .exit
-bash-4.2$ head -20 NodeReport.20161124.151013.8560.001.txt
================================================================================
==== NodeReport ================================================================

Event: JavaScript API, location: "TriggerReport"
Filename: NodeReport.20161124.151013.8560.001.txt
Dump event time:  2016/11/24 15:10:13
Module load time: 2016/11/24 15:10:13
Process ID: 8560

Node.js version: v6.9.1
(v8: 5.1.281.83, libuv: 1.9.1, zlib: 1.2.8, ares: 1.10.1-DEV)

OS version: Linux 3.10.0-327.18.2.el7.x86_64 #1 SMP Fri Apr 8 05:09:53 EDT 2016
(glibc: 2.17)

Machine: drx-hemera x86_64

================================================================================
==== JavaScript Stack Trace ====================================================

-bash-4.2$
```

Looks like compile time constants are being used in [`src/nodereport.cc`](https://github.com/nodejs/nodereport/blob/v1.0.6/src/node_report.cc#L372):
```c++
static void PrintVersionInformation(FILE* fp) {

  // Print Node.js and deps component versions
  fprintf(fp, "\nNode.js version: %s\n", NODE_VERSION);
  fprintf(fp, "(v8: %s, libuv: %s, zlib: %s, ares: %s)\n",
[...]

PR #30, opened 2016-11-24
Fix version reporting in NodeReport section

richardlau:
Fixes: nodejs/nodereport#29

Replaces compile time constants with the contents of `process.versions`. As a bonus `ares.h` and `zlib.h` are no longer required, which should unblock nodejs/citgm#226.

Report content before changes:
```
Node.js version: v6.9.1
(v8: 5.1.281.84, libuv: 1.9.1, zlib: 1.2.8, ares: 1.10.1-DEV)
```
after changes:
```
process.versions = {
  http_parser: 2.7.0
  node: 6.9.1
  v8: 5.1.281.84
  uv: 1.9.1
  zlib: 1.2.8
  ares: 1.10.1-DEV
  icu: 57.1
  modules: 48
  openssl: 1.0.2j
}
```

Tests have been extended to validate the versions contained in the report.

### DIFF
--- a/src/node_report.cc
+++ b/src/node_report.cc
@@ -1,11 +1,9 @@
 #include "node_report.h"
-#include "node_version.h"
 #include "v8.h"
 #include "time.h"
-#include "zlib.h"
-#include "ares.h"
 
 #include <fcntl.h>
+#include <map>
 #include <string.h>
 #include <stdio.h>
 #include <stdlib.h>
@@ -35,6 +33,14 @@
 #include <sys/utsname.h>
 #endif
 
+#if !defined(NODEREPORT_VERSION)
+#define NODEREPORT_VERSION "dev"
+#endif
+
+#if !defined(UNKNOWN_NODEVERSION_STRING)
+#define UNKNOWN_NODEVERSION_STRING "Unable to determine Node.js version\n"
+#endif
+
 #ifndef _WIN32
 extern char** environ;
 #endif
@@ -52,7 +58,7 @@ using v8::String;
 using v8::V8;
 
 // Internal/static function declarations
-static void PrintVersionInformation(FILE* fp);
+static void PrintVersionInformation(FILE* fp, Isolate* isolate);
 static void PrintJavaScriptStack(FILE* fp, Isolate* isolate, DumpEvent event, const char* location);
 static void PrintStackFromStackTrace(FILE* fp, Isolate* isolate, DumpEvent event);
 static void PrintStackFrame(FILE* fp, Isolate* isolate, Local<StackFrame> frame, int index, void* pc);
@@ -70,6 +76,7 @@ const char* v8_states[] = {"JS", "GC", "COMPILER", "OTHER", "EXTERNAL", "IDLE"};
 static bool report_active = false; // recursion protection
 static char report_filename[NR_MAXNAME + 1] = "";
 static char report_directory[NR_MAXPATH + 1] = ""; // defaults to current working directory
+static std::string version_string = UNKNOWN_NODEVERSION_STRING;
 #ifdef _WIN32
 static SYSTEMTIME loadtime_tm_struct; // module load time
 #else  // UNIX, OSX
@@ -187,6 +194,97 @@ unsigned int ProcessNodeReportVerboseSwitch(const char* args) {
     fprintf(stderr, "Unrecognised argument for nodereport verbose switch option: %s\n", args);
   }
   return 0;  // Default is verbose mode off
+}
+
+void SetVersionString(Isolate* isolate) {
+  // Catch anything thrown and gracefully return
+  Nan::TryCatch trycatch;
+  version_string = UNKNOWN_NODEVERSION_STRING;
+
+  // Retrieve the process object
+  v8::Local<v8::String> process_prop;
+  if (!Nan::New<v8::String>("process").ToLocal(&process_prop)) return;
+  v8::Local<v8::Object> global_obj = isolate->GetCurrentContext()->Global();
+  v8::Local<v8::Value> process_value;
+  if (!Nan::Get(global_obj, process_prop).ToLocal(&process_value)) return;
+  if (!process_value->IsObject()) return;
+  v8::Local<v8::Object> process_obj = process_value.As<v8::Object>();
+
+  // Get process.version
+  v8::Local<v8::String> version_prop;
+  if (!Nan::New<v8::String>("version").ToLocal(&version_prop)) return;
+  v8::Local<v8::Value> version;
+  if (!Nan::Get(process_obj, version_prop).ToLocal(&version)) return;
+
+  // e.g. Node.js version: v6.9.1
+  if (version->IsString()) {
+    Nan::Utf8String node_version(version);
+    version_string = "Node.js version: ";
+    version_string += *node_version;
+    version_string += "\n";
+  }
+
+  // Get process.versions
+  v8::Local<v8::String> versions_prop;
+  if (!Nan::New<v8::String>("versions").ToLocal(&versions_prop)) return;
+  v8::Local<v8::Value> versions_value;
+  if (!Nan::Get(process_obj, versions_prop).ToLocal(&versions_value)) return;
+  if (!versions_value->IsObject()) return;
+  v8::Local<v8::Object> versions_obj = versions_value.As<v8::Object>();
+
+  // Get component names and versions from process.versions
+  v8::Local<v8::Array> components;
+  if (!Nan::GetOwnPropertyNames(versions_obj).ToLocal(&components)) return;
+  v8::Local<v8::Object> components_obj = components.As<v8::Object>();
+  std::map<std::string, std::string> comp_versions;
+  uint32_t total_components = (*components)->Length();
+  for (uint32_t i = 0; i < total_components; i++) {
+    v8::Local<v8::Value> name_value;
+    if (!Nan::Get(components_obj, i).ToLocal(&name_value)) continue;
+    v8::Local<v8::Value> version_value;
+    if (!Nan::Get(versions_obj, name_value).ToLocal(&version_value)) continue;
+
+    Nan::Utf8String component_name(name_value);
+    Nan::Utf8String component_version(version_value);
+    if (*component_name == nullptr || *component_version == nullptr) continue;
+
+    // Don't duplicate the Node.js version
+    if (!strcmp("node", *component_name)) {
+      if (version_string == UNKNOWN_NODEVERSION_STRING) {
+        version_string = "Node.js version: v";
+        version_string += *component_version;
+        version_string += "\n";
+      }
+    } else {
+      comp_versions[*component_name] = *component_version;
+    }
+  }
+
+  // Format sorted component versions surrounded by (), wrapped
+  // e.g. 
+  // (ares: 1.10.1-DEV, http_parser: 2.7.0, icu: 57.1, modules: 48, 
+  //  openssl: 1.0.2j, uv: 1.9.1, v8: 5.1.281.84, zlib: 1.2.8)
+  const size_t wrap = 80;
+  size_t line_length = 1;
+  version_string += "(";
+  for (auto it : comp_versions) {
+    std::string component_name = it.first;
+    std::string component_version = it.second;
+    size_t length = component_name.length() + component_version.length() + 2;
+    if (line_length + length + 1 >= wrap) {
+      version_string += "\n ";
+      line_length = 1;
+    }
+    version_string += component_name;
+    version_string += ": ";
+    version_string += component_version;
+    line_length += length;
+    if (it != *std::prev(comp_versions.end())) {
+      version_string += ", ";
+      line_length += 2;
+    }
+  }
+  version_string += ")\n";
 }
 
 /*******************************************************************************
@@ -317,7 +415,7 @@ void TriggerNodeReport(Isolate* isolate, DumpEvent event, const char* message, c
   fflush(fp);
 
   // Print Node.js and OS version information
-  PrintVersionInformation(fp);
+  PrintVersionInformation(fp, isolate);
   fflush(fp);
 
 // Print summary JavaScript stack backtrace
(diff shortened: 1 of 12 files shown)


src/node_report.cc
```diff
@@ -369,12 +366,28 @@ void TriggerNodeReport(Isolate* isolate, DumpEvent event, const char* message, c
...
+      .ToLocalChecked();
+    for (uint32_t i=0, nKeys = (*keys)->Length(); i<nKeys; i++) {
+      Local<Value> keyVal = Nan::Get(keys.As<Object>(), i).ToLocalChecked();
+      Local<Value> valVal = Nan::Get(versionsObj, keyVal).ToLocalChecked();
+      Nan::Utf8String name(keyVal);
+      Nan::Utf8String version(valVal);
+      fprintf(fp, "  %s: %s\n", *name, *version);
+    }
+  }
+  fprintf(fp, "}\n");
```
bnoordhuis: How sure are you that it's safe to call into the VM at this point?

Also, spaces around operators a few lines up?

richardlau: Spaces fixed, thanks.

Other bits of the report later on (e.g. GC statistics, JavaScript stack) are generated via VM calls, so I guess as safe as they are? For this PR an alternative to avoid VM calls here is perhaps capturing `process.versions` into a static buffer on module initialization and then outputting the captured buffer when the report is written.

bnoordhuis: I would capture it at the earliest opportunity, that will let you deal with errors more gracefully.  For example, if user JS code (however unwise) did the below, your code would assert and abort.
```js
Object.defineProperty(process, 'versions', {get() { throw 'boom' }});
```
This particular example is a little farfetched but buggy code clobbering `process` is not unheard of.

rnchamberlain:
I was trying to avoid going into JavaScript-land if at all possible. As Ben says, things may be damaged, especially JS objects on the heap. For v8 there is a nice C++ API - `V8::GetVersion()` - but for Node and the other components just the compile-time constants. I guess building the npm on one Node patch level and running on another is quite likely though, and we do want the actual runtime versions in the report.

Yes we should capture the versions at module load time and stash them rather than access the process object at dump time. 

Incidently, using SetAbortOnUncaughtExceptionCallback() to intercept exceptions is not working very well, so I may need to switch that to use process.on('uncaughtException'). Longer term I think we'll need better APIs in v8 and Node, but that's a long haul.

One other design thing was to make NodeReport easily readable by non-programmers, admins etc. So I prefer something that headlines the Node version like 
`Node.js version: v6.9.1`

rather than 'programmer-style' output
`process.versions = {`
`  http_parser: 2.7.0`
`  node: 6.9.1 ....`

richardlau:
Updated to capture versions at module load time. 

Changed the output to be slightly more human readable, currently:
```
http_parser version: 2.7.0
node version: 6.9.1
v8 version: 5.1.281.84
uv version: 1.9.1
zlib version: 1.2.8
ares version: 1.10.1-DEV
icu version: 57.1
modules version: 48
openssl version: 1.0.2j
```
Comments on formatting? I could pull out `node` so that it's first in the listed output. 

I think I also need to change `SetVersionString()` to check the `MaybeLocal`'s instead of calling `ToLocalChecked()` so that the code can fail gracefully (instead of aborting).

rnchamberlain:
Yes, suggest have node and v8 top of the list as they are of most interest to people I think. I had the less interesting components on a single line, to keep the intro section short - you don't have to scroll down far to see the stack traces, but it's not a big deal. 

Anything to make the code resilient - so we carry on - is good, with some 'version information not available' indication maybe. Thanks!

sam-github:
Wouldn't it be more consistent to have output be

````
====================================
==== Versions ========================

node: xxx
openssl: xxx
...
```

I would tend to sort the components lexically.

sam-github:
Ah, scratch that last comment, I'm sorry. I liked the original format, but all the process.versions keys should
be mentioned, with the same key! `libuv` -> `uv`. And I'd sort.

```
Node.js version: v6.9.1
(ares: 1.10.1-DEV, http_parser: 2.7.0, icu: 57.1, modules: 48, openssl: 1.0.2j, uv: 1.9.1, v8: 5.1.281.84, zlib: 1.2.8)
```

As an aside, `modules` is an absolutely terrible name for the ABI version, and its meaning is undocumented,  I just had to read the source to figure out what it was (since it clearly isn't a component in `deps/`, like everything else except node. :-(. /cc @mhdawson

richardlau:
My issue with including all the components on one line is that the line is quite long. 

Should we filter out `modules` from the list?

sam-github:
No, patching over oddities in node just makes things odder, we'd have two special snowflakes, not one.

The line is not longer than lots of other lines, so I don't think its a big issue, but I see your point. I  think repeating the word "version" in every line is ugly and repetitive. Perhaps:

```
Node.js version: v6.9.1
  ares: 1.10.1-DEV
  http_parser: 2.7.0
  icu: 57.1
  modules: 48
  openssl: 1.0.2j
  uv: 1.9.1
  v8: 5.1.281.84
  zlib: 1.2.8
```

mhdawson:
I think people understand what

mhdawson:
@sam-github I agree that modules: is not a great name, but thinking we want to be consistent with process.versions.

richardlau:
Updated so the output is now similar to what it was before. I've enclosed the components (i.e. everything in `process.versions` except for `node`) in `()` and wrapped at 80 characters. I aligned subsequent lines (i.e. the second line in the sample output is indented by one ` ` character) but that can easily be changed. 
```
Node.js version: v6.9.1
(ares: 1.10.1-DEV, http_parser: 2.7.0, icu: 57.1, modules: 48, openssl: 1.0.2j, 
 uv: 1.9.1, v8: 5.1.281.84, zlib: 1.2.8)
```
I left the tests so that they only check the information is there (and at the moment ignore formatting). 

I'll squash later on.

While I'm editing in this area, as per nodejs/nodereport#29 I'll add a line to this section of the report that reports the version of nodereport used and which version of Node.js it was compiled against, e.g. `nodereport v1.0.6 built against Node.js v6.9.1`.

richardlau:
Oh I also attempted to make the code resilient, with a default `Unable to determine Node.js version` if something goes wrong. I'll try to add a test to verify this (probably by clobbering `process.version`/`process.versions`.

richardlau:
Possibly stupid question, should it be 
```
nodereport v1.0.6 built against Node.js v6.9.1
```
or
```
NodeReport v1.0.6 built against Node.js v6.9.1
```
?

sam-github:
`NodeReport v1.0.6 built against Node.js v6.9.1` or `nodereport v1.0.6 built against node v6.9.1`, IMO.

mhdawson:
It should be Node.js, not so sure about nodereport versus NodeReport but I'm leaning to NodeReport.

src/node_report.cc
```diff
@@ -70,6 +68,7 @@ const char* v8_states[] = {"JS", "GC", "COMPILER", "OTHER", "EXTERNAL", "IDLE"};
 static bool report_active = false; // recursion protection
 static char report_filename[NR_MAXNAME + 1] = "";
 static char report_directory[NR_MAXPATH + 1] = ""; // defaults to current working directory
+static std::string versionString = "Unable to determine Node.js version\n";
```
bnoordhuis: Can you use snake_case for consistency?  (Here and elsewhere, e.g., also please update processProp, versionProp, and so on.)

src/node_report.cc
```diff
@@ -189,6 +188,109 @@ unsigned int ProcessNodeReportVerboseSwitch(const char* args) {
...
+  Nan::MaybeLocal<String> processProp = Nan::New<String>("process");
+  if (processProp.IsEmpty() || trycatch.HasCaught()) {
+    trycatch.Reset();
+    return;
+  }
+  Nan::MaybeLocal<Value> process = Nan::Get(
+    isolate->GetCurrentContext()->Global(),
+    processProp.ToLocalChecked());
+  if (process.IsEmpty() || process.ToLocalChecked()->IsUndefined()
+      || trycatch.HasCaught()) {
```
bnoordhuis: I don't know how much core style is supposed to be followed but assuming it's the same style:

* the operator should go on the previous line
* line continuations in the Nan::Get() call should be indented with four spaces

The call to IsUndefined() looks like it should be `!process.ToLocalChecked()->IsObject()`?

src/node_report.cc
```diff
@@ -189,6 +188,109 @@ unsigned int ProcessNodeReportVerboseSwitch(const char* args) {
   return 0;  // Default is verbose mode off
 }
 
+void SetVersionString(Isolate* isolate) {
+  Nan::TryCatch trycatch;
+  Nan::MaybeLocal<String> processProp = Nan::New<String>("process");
+  if (processProp.IsEmpty() || trycatch.HasCaught()) {
+    trycatch.Reset();
```
bnoordhuis: .Reset() is the default behaviour unless .ReThrow() is called, you don't need to call it explicitly.  You also don't really need to check for .HasCaught() because .IsEmpty() is always true in case of exception.

A more succinct way of writing it:
```c++
Local<String> process_prop;
if (!Nan::New<String>("process").ToLocal(&process_prop)) return;
```

src/node_report.cc
```diff
@@ -189,6 +188,109 @@ unsigned int ProcessNodeReportVerboseSwitch(const char* args) {
...
+    compString.append(*compVersion);
+    compVersions.push_back(compString);
+  }
+
+  // Log sorted component versions in (), wrapped
+  const size_t wrap = 80;
+  size_t lineLength = 1;
+  versionString.append("(");
+  std::sort(compVersions.begin(), compVersions.end());
+  for (std::vector<std::string>::iterator it = compVersions.begin();
```
bnoordhuis: auto it?

richardlau: done

src/node_report.cc
richardlau: done

richardlau:
Fixed suggestions from @bnoordhuis.

Added reporting of NodeReport version and test.
```
Node.js version: v6.9.1
(ares: 1.10.1-DEV, http_parser: 2.7.0, icu: 57.1, modules: 48, openssl: 1.0.2j, 
 uv: 1.9.1, v8: 5.1.281.84, zlib: 1.2.8)

NodeReport version: 1.0.6 (built against Node.js v6.9.1)
```

richardlau:
Made the code even more resilient -- It will now get the Node.js version from `process.versions` if `process.version` is unavailable. If neither `process.version` nor `process.versions` is available, the report will contain:
```
Unable to determine Node.js version

NodeReport version: 1.0.6 (built against Node.js v6.9.1)
```

Squashed commits.

rnchamberlain:
LGTM, new tests all ran ok (on Linux)

@bnoordhuis - are your review comments all resolved now? thanks

GeorgeAdams95:
@bnoordhuis is this good to merge because this is blocking us from adding nodereport to citgm?

src/node_report.cc
```diff
@@ -189,6 +196,95 @@ unsigned int ProcessNodeReportVerboseSwitch(const char* args) {
...
+  Nan::TryCatch trycatch;
+  version_string = UNKNOWN_NODEVERSION_STRING;
+
+  // Retrieve the process object
+  v8::Local<v8::String> process_prop;
+  if (!Nan::New<v8::String>("process").ToLocal(&process_prop)) return;
+  v8::Local<v8::Object> global_obj = isolate->GetCurrentContext()->Global();
+  v8::Local<v8::Value> process_value;
+  if (!Nan::Get(global_obj, process_prop).ToLocal(&process_value)) return;
+  v8::Local<v8::Object> process_obj = process_value.As<v8::Object>();
```
bnoordhuis: For the sake of robustness, you should `if (!process_value->IsObject()) return;` first.

src/node_report.cc
```diff
@@ -189,6 +196,95 @@ unsigned int ProcessNodeReportVerboseSwitch(const char* args) {
...
+  std::map<std::string, std::string> comp_versions;
+  uint32_t total_components = (*components)->Length();
+  for (uint32_t i = 0; i < total_components; i++) {
+    v8::Local<v8::Value> name_value;
+    if (!Nan::Get(components_obj, i).ToLocal(&name_value)) continue;
+    v8::Local<v8::Value> version_value;
+    if (!Nan::Get(versions_obj, name_value).ToLocal(&version_value)) continue;
+
+    Nan::Utf8String component_name(name_value);
+    Nan::Utf8String component_version(version_value);
```
bnoordhuis: Robustness: `if (*component_name == nullptr || *component_version == nullptr) continue;`

test/common.js
```diff
@@ -25,20 +27,60 @@ exports.isWindows = () => {
...
+        t.match(nodeReportSection, /Unable to determine Node.js version/,
+                'NodeReport section contains expected Node.js version');
+      } else {
+        t.match(nodeReportSection,
+                new RegExp('Node.js version: ' + process.version),
+                'NodeReport section contains expected Node.js version');
+      }
+      nodeComponents.forEach((c) => {
+        if (c !== 'node') {
+          if (expectedVersions.indexOf(c) === -1) {
```
bnoordhuis: Might as well use .includes() here.

richardlau: NodeReport claims to be Node.js v4.x compatible (`"node": ">=4.0.0"`) and the version of V8 in v4.x doesn't have `Array.prototype.includes()`.

src/node_report.cc
richardlau: Done

src/node_report.cc
richardlau: Done

richardlau:
Updated to address Ben's robustness comments. Haven't implemented the suggested use of `.includes()` as that isn't supported in Node.js v4.x (which NodeReport supports).

rnchamberlain:
Landed as https://github.com/nodejs/nodereport/commit/368f253ea7c40f72e11319261e9d3b58c4501cf7

Issue #22, opened 2016-11-15
clang warning when using MacOS Sierra


AndyRWatson:
when attempting to install on Node v6.7.0 via npm using MacOS 10.12.1 (Sierra)  I get the following compiler warning

clang: warning: libstdc++ is deprecated; move to libc++ with a minimum deployment target of OS X 10.9

gibfahn:
Node core issue here: https://github.com/nodejs/node/issues/8521

rnchamberlain:
Possible work-around here: https://github.com/nodejs/nan/issues/606

rnchamberlain:
Fix landed in github, not published in npm yet.

JeremyIglehart:
This is all still way over my head - I'm still experiencing this.  Any ETA when this will make it to NPM?

rnchamberlain:
Should be on npm in the next day or so. Quite a few changes have gone in and it's a new version, just need to do the project rename, see https://github.com/nodejs/nodereport/issues/35 https://github.com/nodejs/nodereport/pull/43

JeremyIglehart:
Wow.  Amazing work has been done here.  You all work together better than most companies do!  Hat's off to everyone on this project.  It was encouraging to read through those issues.  Thank you for al the hard work you're all working on.

richardlau:
We've decided in #56 to revert the change made by #31 and instead be consistent with Node.js and link to the same library as it does. That means that if you build with Node.js v6 you will get deprecation warnings (you will get the same deprecation warning if you build Node.js v6, see https://github.com/nodejs/node/issues/8521#issuecomment-260649919).

Node.js v7 and above have switched to using libc++, so you will not get the warnings with those.

rnchamberlain:
Closing this as per @richardlau comment above, fixed in Node 7 and later.

PR #31, opened 2016-12-08
Fix for clang warning: libstdc++ is deprecated

rnchamberlain:
When installing the nodereport module on macOS 10.12 (Sierra), the clang deprecation warning above is output during npm install. The fix is to override the std c++ library inherited from node, so that the module is linked to libc++

Uses known solution described here: https://github.com/nodejs/nan/issues/606

Fixes https://github.com/nodejs/nodereport/issues/22

### DIFF
--- a/binding.gyp
+++ b/binding.gyp
@@ -13,6 +13,12 @@
           "libraries": [ "dbghelp.lib", "Netapi32.lib" ],
           "dll_files": [ "dbghelp.dll", "Netapi32.dll" ],
         }],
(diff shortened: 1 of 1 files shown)


gibfahn:
Seems to fix https://github.com/nodejs/nodereport/issues/32

rnchamberlain:
Landed as https://github.com/nodejs/nodereport/commit/67050f527809080e78dd399b6a8e48e369b5f486

PR #33, opened 2016-12-14
Opt-in hooks and signal by default

rnchamberlain:
This changes the default action on require('nodereport') to opt-in to the exception and fatal-error hooks, and to enable the signal handler. So 'node -r nodereport application.js' will get all the nodereport functionality without any application code changes.

For users programming to the API, a separate api-only entry point, require('nodereport/api') is provided. This does not enable the hooks and signal unless and until .setEvents() is called.

Addresses issue https://github.com/nodejs/nodereport/issues/25

This is a follow-on from PR https://github.com/nodejs/nodereport/pull/28

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,0 +1,20 @@
+// Main module entry point for nodereport
+
+const api = require('./api');
+
+// Process NODEREPORT_EVENTS env var
+const options = process.env.NODEREPORT_EVENTS;
+if (options) {
+  api.setEvents(options);
+} else {
+  // Default action - all events enabled
+  api.setEvents('exception+fatalerror+signal+apicall');
+}
+
+exports.triggerReport = api.triggerReport;
+exports.setEvents = api.setEvents;
+exports.setCoreDump = api.setCoreDump;
+exports.setSignal = api.setSignal;
+exports.setFileName = api.setFileName;
+exports.setDirectory = api.setDirectory;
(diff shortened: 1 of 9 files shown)


rnchamberlain:
@sam-github @richardlau @Fishrock123 this should be the same as the final version in the earlier PR https://github.com/nodejs/nodereport/pull/28 but a quick re-review would be much appreciated - thanks

index.js
```diff
@@ -0,0 +1,20 @@
+// Main module entry point for nodereport
+
+const api = require('./api');
+
+// Process NODEREPORT_EVENTS env var
+const options = process.env.NODEREPORT_EVENTS;
```
sam-github: You can simplify and remove the conditional if you want:
```
const options = process.env.NODEREPORT_EVENTS || 'exception+fatalerror+signal+apicall';
api.setEvents(options);
```

rnchamberlain:
@sam-github agree, needs a positive test for `require('nodereport/api') `with a call to `nodereport.triggerReport()` I'll clone test-api.js as test-api-nohooks.js and then we'll have coverage for both routes.

index.js
rnchamberlain: nice, done

rnchamberlain:
Landed as https://github.com/nodejs/nodereport/commit/8605e4cae6bf6125cffa856027caf26f90b1e6ac

PR #34, opened 2016-12-16
Add command line to nodereport.

hhellyer:
This patch adds the command line used to start the process to nodereport.
It's based off https://github.com/rnchamberlain/nodereport/tree/clang_warning so it could be built and tested on Mac. It should merge cleanly once that has been landed.

One line is added to the NodeReport file. Sample output from a test script that generates a report:

`> node --no-warnings my_application.js -a an_argument file.txt`

First part of the generated NodeReport file:
```
==== NodeReport ================================================================

Event: JavaScript API, location: "TriggerReport"
Filename: NodeReport.20161216.145623.6353.001.txt
Dump event time:  2016/12/16 14:56:23
Module load time: 2016/12/16 14:56:23
Process ID: 6353
Command line arguments: node --no-warnings my_application.js -a an_argument file.txt

Node.js version: v6.5.0
```

### DIFF
--- a/src/node_report.cc
+++ b/src/node_report.cc
@@ -313,37 +313,37 @@ void SetLoadTime() {
  * Function to save the process command line
  *******************************************************************************/
 void SetCommandLine() {
-  #ifdef __linux__
-    // Read the command line from /proc/self/cmdline
-    char buf[64];
-    FILE* cmdline_fd = fopen("/proc/self/cmdline", "r");
-    if (cmdline_fd == nullptr) {
-      return;
-    }
-    commandline_string = "";
-    int bytesread = fread(buf, 1, sizeof(buf), cmdline_fd);
-    while (bytesread > 0) {
-      for (int i = 0; i < bytesread; i++) {
-        // Arguments are null separated.
-        if (buf[i] == '\0') {
-          commandline_string += " ";
-        } else {
-          commandline_string += buf[i];
-        }
+#ifdef __linux__
+  // Read the command line from /proc/self/cmdline
+  char buf[64];
+  FILE* cmdline_fd = fopen("/proc/self/cmdline", "r");
+  if (cmdline_fd == nullptr) {
+    return;
+  }
+  commandline_string = "";
+  int bytesread = fread(buf, 1, sizeof(buf), cmdline_fd);
+  while (bytesread > 0) {
+    for (int i = 0; i < bytesread; i++) {
+      // Arguments are null separated.
+      if (buf[i] == '\0') {
+        commandline_string += " ";
+      } else {
+        commandline_string += buf[i];
       }
-      bytesread = fread(buf, 1, sizeof(buf), cmdline_fd);
     }
-    fclose(cmdline_fd);
-  #elif __APPLE__
-    char **argv = *_NSGetArgv();
-    int argc = *_NSGetArgc();
+    bytesread = fread(buf, 1, sizeof(buf), cmdline_fd);
+  }
+  fclose(cmdline_fd);
+#elif __APPLE__
+  char **argv = *_NSGetArgv();
+  int argc = *_NSGetArgc();
 
-    commandline_string = "";
-    std::string seperator = "";
-    for (int i = 0; i < argc; i++) {
-      commandline_string += seperator + argv[i];
-      seperator = " ";
-    }
+  commandline_string = "";
+  std::string seperator = "";
+  for (int i = 0; i < argc; i++) {
+    commandline_string += seperator + argv[i];
+    seperator = " ";
+  }
 #elif _AIX
   // Read the command line from /proc/self/cmdline
   char procbuf[64];
--- a/src/node_report.cc
+++ b/src/node_report.cc
@@ -364,7 +364,9 @@ void SetCommandLine() {
       separator = " ";
     }
   }
-#endif // _AIX
+#elif _WIN32
+  commandline_string = GetCommandLine();
+#endif
 }
 
 /*******************************************************************************
@@ -541,7 +543,7 @@ void TriggerNodeReport(Isolate* isolate, DumpEvent event, const char* message, c
  ******************************************************************************/
 static void PrintCommandLine(FILE* fp) {
   if (commandline_string != "") {
-    fprintf(fp, "Command line arguments: %s\n", commandline_string.c_str());
+    fprintf(fp, "Command line: %s\n", commandline_string.c_str());
   }
 }
 
--- a/src/node_report.cc
+++ b/src/node_report.cc
@@ -339,10 +339,10 @@ void SetCommandLine() {
   int argc = *_NSGetArgc();
 
   commandline_string = "";
-  std::string seperator = "";
+  std::string separator = "";
   for (int i = 0; i < argc; i++) {
-    commandline_string += seperator + argv[i];
-    seperator = " ";
+    commandline_string += separator + argv[i];
+    separator = " ";
   }
 #elif _AIX
   // Read the command line from /proc/self/cmdline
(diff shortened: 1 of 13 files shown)


Fishrock123:
@hhellyer could you document how this is supposed to be used? :D

hhellyer:
I've added some sample output.

cjihrig:
@hhellyer it looks like there are conflicts in `src/node_report.cc` that need to be resolved.

richardlau:
Is it possible to get the equivalent information on AIX and Windows?

richardlau:
Also needs tests.

hhellyer:
@richardlau 
Not sure about AIX, it doesn't have /proc/self/cmdline so I'll need to find something else.

It looked like Windows had globals (__argv, __argc) which you could use, though I've not tested them as I don't have a windows box easily accessible. See: https://msdn.microsoft.com/en-us/library/dn727674.aspx for details, there may be a better way - I only had a quick look.

richardlau:
Windows: [GetCommandLine()](https://msdn.microsoft.com/en-gb/library/windows/desktop/ms683156%28v=vs.85%29.aspx)

AIX: [getargs()](http://www.ibm.com/support/knowledgecenter/ssw_aix_61/com.ibm.aix.basetrf1/getargs.htm). I actually used this to fix libuv/libuv#464 (although in that case we were only interested in the first argument of the command line).

rnchamberlain:
Looks good
1. Suggest just `Command line: `rather than `Command line arguments: `
2. Windows and AIX support would be great - if not too much work
3. Not sure we'll want tests for every output line in NodeReport, but as this is using some significant OS platform calls then yes maybe a test would be good.
4. I've landed the Mac build fix, so probably best to rebase this now

hhellyer:
I've added AIX support. Note that on AIX we get the wrong command but that's because of https://github.com/nodejs/node/issues/10607 and we give the same result as the ps command.

I've squashed the commits down, it looks like github shows the changes on the branch and not a diff against the merge destination which is confusing if you pull in changes from there, which I had done. At least I think that's what was happening... :-)

gibfahn:
@hhellyer AFAIK the [Files](https://github.com/nodejs/nodereport/pull/34/files) is just a diff against the merge destination, and if you want to see individual commit diffs you look at [Commits](https://github.com/nodejs/nodereport/pull/34/commits). However Github does weird things sometimes.

src/node_report.cc
```diff
@@ -299,6 +308,65 @@ void SetLoadTime() {
...
+    char **argv = *_NSGetArgv();
+    int argc = *_NSGetArgc();
+
+    commandline_string = "";
+    std::string seperator = "";
+    for (int i = 0; i < argc; i++) {
+      commandline_string += seperator + argv[i];
+      seperator = " ";
+    }
+#elif _AIX
```
hhellyer: The rest of the function was wrong and didn't match the file. I'd indented when starting #ifdef sections. I've adjusted the rest of the function to be correct.

src/node_report.cc
```diff
@@ -299,6 +308,65 @@ void SetLoadTime() {
...
+    }
+    bytesread = fread(buf, 1, sizeof(buf), cmdline_fd);
+  }
+  fclose(cmdline_fd);
+#elif __APPLE__
+  char **argv = *_NSGetArgv();
+  int argc = *_NSGetArgc();
+
+  commandline_string = "";
+  std::string seperator = "";
```
richardlau: nit: `seperator` -> `separator`

src/node_report.cc
```diff
@@ -299,6 +308,65 @@ void SetLoadTime() {
...
+  FILE* psinfo_fd = fopen(procbuf, "r");
+  if (psinfo_fd == nullptr) {
+    return;
+  }
+  psinfo_t info;
+  int bytesread = fread(&info, 1, sizeof(psinfo_t), psinfo_fd);
+  fclose(psinfo_fd);
+  if (bytesread == sizeof(psinfo_t)) {
+    commandline_string = "";
+    std::string seperator = "";
```
richardlau: nit: `seperator` -> `separator`

hhellyer:
I've added the testcases or extended the existing ones.
(I've also fixed the spelling mistake.)

rnchamberlain:
Added commit https://github.com/nodejs/nodereport/pull/34/commits/dd201873a3ff9d553512dfee0f3fcb53cf8c6f45 for support on Windows. I think this PR has all we need now.

test/common.js
```diff
@@ -36,18 +36,18 @@ exports.validate = (t, report, options) => {
       const expectedVersions = options ?
                                options.expectedVersions || nodeComponents :
                                nodeComponents;
-      const plan = REPORT_SECTIONS.length + nodeComponents.length + 2;
+      var plan = REPORT_SECTIONS.length + nodeComponents.length + 2;
+      if( options.commandline ) plan++;
```
richardlau: Spaces -> `if (options.commandline) plan++;`

test/common.js
```diff
@@ -36,18 +36,18 @@ exports.validate = (t, report, options) => {
...
       REPORT_SECTIONS.forEach((section) => {
         t.match(reportContents, new RegExp('==== ' + section),
                 'Checking report contains ' + section + ' section');
       });
 
-      // Check NodeReport section
+      // Check NodeReport header section
       const nodeReportSection = getSection(reportContents, 'NodeReport');
-      t.match(nodeReportSection, new RegExp('Process ID: ' + pid),
+      t.contains(nodeReportSection, new RegExp('Process ID: ' + pid),
```
richardlau: Why was this changed?

test/common.js
```diff
@@ -57,6 +57,19 @@ exports.validate = (t, report, options) => {
                 new RegExp('Node.js version: ' + process.version),
                 'NodeReport section contains expected Node.js version');
       }
+      if (options && options.commandline) {
+        if (this.isWindows()) {
+          // On Windows we need to strip out double quotes from the command line in the
+          // report, and escape the backslashes in the regex comparison string.
+          t.match(nodeReportSection.replace(/"/g,''),
+                  new RegExp('Command line: ' + (options.commandline).replace(/\\/g,'\\\\')),
```
richardlau: I think this line is > 80 chars.

test/common.js
rnchamberlain: Fixed

test/common.js
rnchamberlain: @hhellyer said it fixed a test failure, but http://www.node-tap.org/asserts/ indicatess that .contains is simply a synonym for .match, so I have reverted it. The tests pass ok for me on Linux, Windows and Mac.

test/common.js
rnchamberlain: Fixed

test/common.js
```diff
@@ -57,6 +57,20 @@ exports.validate = (t, report, options) => {
                 new RegExp('Node.js version: ' + process.version),
                 'NodeReport section contains expected Node.js version');
       }
+      if (options && options.commandline) {
+        if (this.isWindows()) {
+          // On Windows we need to strip double quotes from the command line in
+          // the report, and escape backslashes in the regex comparison string.
```
richardlau: n.b. I'm not seeing any double quotes in the reported command line in the reports generated from the tests on Windows.

rnchamberlain: It might depend on the shell or the executable path. This is what I am seeing on Windows:
```
==== NodeReport ================================================================

Event: JavaScript API, location: "TriggerReport"
Filename: NodeReport.20170110.113318.2184.001.txt
Dump event time:  2017/01/10 11:33:18
Module load time: 2017/01/10 11:33:18
Process ID: 2184
Command line: "C:\Program Files\nodejs6\node.exe" C:\tmp\nodereport\test\test-api-bad-processobj.js child
```

rnchamberlain:
Landed as https://github.com/nodejs/nodereport/commit/c6db0733f1c789de9ecfa57a2dbd2c17b8e32b1f

test/common.js
richardlau: Ah I think it might be spaces in the paths -- I get the same if I use Node.js installed under `C:\Program Files`

PR #36, opened 2017-01-12
Fix behaviour on exception to match node default

rnchamberlain:
This patch switches nodereport behaviour on uncaught exceptions to match default Node behaviour.  Previously nodereport was triggering an abort/core dump unless the .setCoreDump() API or the NODEREPORT_COREDUMP env var were used to switch that off.

We now detect whether the user specified the -abort-on-uncaught-exception option and honor that.

Fixes https://github.com/nodejs/nodereport/issues/6

The .setCoreDump() API and NODEREPORT_COREDUMP env var are now redundant and are removed. So this will be a semver major change - and go with proposed rename to "node-report".

Note: on Node v4 and v6, if false is returned from the callback set via isolate->SetAbortOnUncaughtExceptionCallback() - ie. no abort is requested - Node terminates silently. On Node v7 it prints the usual exception message and stack trace. As a workaround for this on v4 and v6, the nodereport code will print the exception stack trace.

### DIFF
--- a/src/module.cc
+++ b/src/module.cc
@@ -160,7 +160,7 @@ bool OnUncaughtException(v8::Isolate* isolate) {
   // to stderr, as V8 does not do so (unless we trigger an abort, as above).
   int v8_major, v8_minor;
   if (sscanf(v8::V8::GetVersion(), "%d.%d", &v8_major, &v8_minor) == 2) {
-    if (v8_major <= 5 && v8_minor < 4) {
+    if (v8_major < 5 || (v8_major == 5 && v8_minor < 4)) {
       fprintf(stderr, "\nUncaught exception at:\n");
 #ifdef _WIN32
       // On Windows, print the stack using StackTrace API
--- a/src/module.cc
+++ b/src/module.cc
@@ -6,7 +6,9 @@ namespace nodereport {
 // Internal/static function declarations
 static void OnFatalError(const char* location, const char* message);
 bool OnUncaughtException(v8::Isolate* isolate);
-#ifndef _WIN32
+#ifdef _WIN32
+static void PrintStackFromStackTrace(Isolate* isolate, FILE* fp);
+#else  // signal trigger functions for Unix platforms and OSX
 static void SignalDumpAsyncCallback(uv_async_t* handle);
 inline void* ReportSignalThreadMain(void* unused);
 static int StartWatchdogThread(void* (*thread_main)(void* unused));
@@ -19,7 +21,6 @@ static void SetupSignalHandler();
 
 // Default nodereport option settings
 static unsigned int nodereport_events = NR_APICALL;
-static unsigned int nodereport_core = 1;
 static unsigned int nodereport_verbose = 0;
 #ifdef _WIN32  // signal trigger not supported on Windows
 static unsigned int nodereport_signal = 0;
@@ -28,7 +29,7 @@ static unsigned int nodereport_signal = SIGUSR2; // default signal is SIGUSR2
 static int report_signal = 0;  // atomic for signal handling in progress
 static uv_sem_t report_semaphore;  // semaphore for hand-off to watchdog
 static uv_async_t nodereport_trigger_async;  // async handle for event loop
-static uv_mutex_t node_isolate_mutex;  // mutex for wachdog thread
+static uv_mutex_t node_isolate_mutex;  // mutex for watchdog thread
 static struct sigaction saved_sa;  // saved signal action
 #endif
 
@@ -38,6 +39,8 @@ static bool error_hook_initialised = false;
 static bool signal_thread_initialised = false;
 
 static v8::Isolate* node_isolate;
+extern std::string version_string;
+extern std::string commandline_string;
 
 /*******************************************************************************
  * External JavaScript API for triggering a NodeReport
@@ -100,10 +103,6 @@ NAN_METHOD(SetEvents) {
   }
 #endif
 }
-NAN_METHOD(SetCoreDump) {
-  Nan::Utf8String parameter(info[0]);
-  nodereport_core = ProcessNodeReportCoreSwitch(*parameter);
-}
 NAN_METHOD(SetSignal) {
 #ifndef _WIN32
   Nan::Utf8String parameter(info[0]);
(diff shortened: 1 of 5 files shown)


src/module.cc
```diff
@@ -145,22 +144,63 @@ static void OnFatalError(const char* location, const char* message) {
...
+        fprintf(fp, "at [eval]:%i:%i\n", line_number, column);
+      } else {
+        fprintf(fp, "at [eval] (%s:%i:%i)\n", *script_name, line_number, column);
+      }
+    } else {
+      if (fn_name_s.length() == 0) {
+        fprintf(fp, "%s:%i:%i\n", *script_name, line_number, column);
+      } else {
+        fprintf(fp, "%s (%s:%i:%i)\n", *fn_name_s, *script_name, line_number, column);
+      }
```
richardlau: `PrintStackFrame` in `node_report.cc` also checks `frame->IsConstructor()` -- Does this need to do the same?

test/test-exception.js
```diff
@@ -21,14 +21,8 @@ if (process.argv[2] === 'child') {
...
-  child.on('exit', (code, signal) => {
-    const expectedExitCode = common.isWindows() ? 0xC0000005 : null;
-    const expectedSignal = common.isWindows() ? null :
-                           common.isPPC() ? 'SIGTRAP' : 'SIGILL';
-    tap.plan(4);
-    tap.equal(code, expectedExitCode, 'Process should not exit cleanly');
-    tap.equal(signal, expectedSignal,
-              'Process should exit with expected signal ');
+  child.on('exit', (code) => {
+    tap.plan(2);
```
richardlau: We're no longer checking the expected exit code?

src/module.cc
```diff
@@ -145,22 +144,63 @@ static void OnFatalError(const char* location, const char* message) {
...
-  return nodereport_core;
+  }
+  if ((commandline_string.find("abort-on-uncaught-exception") != std::string::npos) ||
+      (commandline_string.find("abort_on_uncaught_exception") != std::string::npos)) {
+    return true;  // abort required
+  }
+  // On Node v4 and v6 we need to print the exception backtrace to stderr, as
+  // Node does not do so (unless we trigger an abort, as above)
+  if ((version_string.find("v4") != std::string::npos) ||
+      (version_string.find("v6") != std::string::npos)) {
```
rnchamberlain: Yes that is a bit fragile. Maybe I should limit the search to the first part of the string, or search for "Node.js version: v4" etc. Alternatively I could save the Node version in a separate dedicated variable during the parsing in SetVersionString(), and use that instead.

richardlau: Is the behavior dependent on Node.js or is it V8 behavior? Could the check be done using `V8::GetVersion()`?

rnchamberlain: Good point, it is actually V8 behaviour, so I have changed the check to use V8::GetVersion().

src/module.cc
```diff
@@ -145,22 +144,63 @@ static void OnFatalError(const char* location, const char* message) {
...
+    // On other platforms use the Message API
+    Message::PrintCurrentStackTrace(isolate, stderr);
+#endif
+  }
+  return false;
 }
 
-#ifndef _WIN32
+#ifdef _WIN32
+static void PrintStackFromStackTrace(Isolate* isolate, FILE* fp) {
```
rnchamberlain: It's a simplified version of the function in node-report.cc. The stack trace in a NodeReport has quite a bit more detail, e.g. the pc value on each line and the JavaScript state value. I did consider sharing the code but I think the resulting single function would have been more complicated. The order of the parameters matches Message::PrintCurrentStackTrace() which is called on the next line in OnUncaughtException()

src/module.cc
rnchamberlain: As above, the stack trace in a NodeReport file identifies constructors, but that is not needed in the simpler version we are outputting to stderr.

test/test-exception.js
rnchamberlain: Good point, I have added a check for the expected exit code, also a basic check that the top function in the stack trace was there.

rnchamberlain:
Added commit in response to the @richardlau questions

src/module.cc
```diff
@@ -156,18 +156,20 @@ bool OnUncaughtException(v8::Isolate* isolate) {
...
-  // On Node v4 and v6 we need to print the exception backtrace to stderr, as
-  // Node does not do so (unless we trigger an abort, as above)
-  if ((version_string.find("v4") != std::string::npos) ||
-      (version_string.find("v6") != std::string::npos)) {
-    fprintf(stderr, "\nUncaught exception at:\n");
+  // On V8 versions earlier than 5.4 we need to print the exception backtrace
+  // to stderr, as V8 does not do so (unless we trigger an abort, as above).
+  int v8_major, v8_minor;
+  if (sscanf(v8::V8::GetVersion(), "%d.%d", &v8_major, &v8_minor) == 2) {
+    if (v8_major <= 5 && v8_minor < 4) {
```
richardlau: Based on the preceding comment, this check doesn't look correct, e.g. v8 v4.5 (`v8_major == 4`, `v8_minor == 5`) would fail the check.

test/test-exception.js
```diff
@@ -5,24 +5,33 @@ if (process.argv[2] === 'child') {
...
+  // Capture stderr output from the child process
+  var stderr = '';
+  child.stderr.on('data', (chunk) => {
+    stderr += chunk;
+  });
   child.on('exit', (code) => {
-    tap.plan(2);
+    tap.plan(4);
+    const v8_version = (process.versions.v8).match(/\d+/g);
+    if (v8_version[0] <= 5 && v8_version[1] < 4) {
```
richardlau: Same as for the equivalent check in the C++ code -- v8 v4.5 would fail this check.

src/module.cc
rnchamberlain: Doh, yes that check is wrong, fixed now, thanks.

test/test-exception.js
rnchamberlain: Fixed here as well

rnchamberlain:
Landed as https://github.com/nodejs/nodereport/commit/1930ce10bd433ec2b92bb2d2653afe918f8c32d1

Issue #19, opened 2016-11-07
can we report something about the modules in use?


sam-github:
A full `npm ls` tree would be too much, but would there be value in reporting at least any binary addons that have been loaded? I guess if addons fault they will show up in the stack traces, but they may also be sources of more subtle corruption (writes to stray pointers corrupting heap, for example).

rnchamberlain:
Yes this would be a useful enhancement. We could do something a bit more general, and list all binaries from /proc/pid/maps, e.g.:

```
000000400000-000001bbe000 r-xp 08:01 2886147    /home/user/node-v6.9.1-linux-x64/bin/node
7fda2dfc4000-7fda2dfcf000 r-xp 08:01 3015273    /home/user/test/node_modules/nodereport/build/Release/nodereport.node
7fda301d4000-7fda30393000 r-xp 08:01 3281614    /lib/x86_64-linux-gnu/libc-2.23.so
7fda3059d000-7fda305b5000 r-xp 08:01 3281610    /lib/x86_64-linux-gnu/libpthread-2.23.so
7fda307ba000-7fda307d0000 r-xp 08:01 3281420    /lib/x86_64-linux-gnu/libgcc_s.so.1
7fda30cd9000-7fda30e4b000 r-xp 08:01 529796     /usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.21
7fda3105b000-7fda31062000 r-xp 08:01 3281577    /lib/x86_64-linux-gnu/librt-2.23.so
```

mhdawson:
+1 from me on this.

hhellyer:
I've added PR #38 to try to address this, it asks the OS for the list of loaded modules/libraries rather than just looking at the modules node has loaded.

rnchamberlain:
@sam-github this is now available in node-report v2.0.0, the list of loaded libraries appears at the end of the report

sam-github:
Great

PR #38, opened 2017-01-18
Add the list of library files loaded by the process to nodereport.

hhellyer:
This pull request is for issue https://github.com/nodejs/nodereport/issues/19 and adds an extra section to the System Information nodereport provides that shows the libraries loaded by the application. This allows the user to identify any native libraries loaded by node modules as well as any other system libraries that have been loaded by node.

It provides the information reported by the operating system. This is similar to the output of ldd however it includes libraries loaded at runtime rather than just those that the executable is linked against.

This first implementation supports Linux, Mac and AIX. I will ask @rnchamberlain if he can push an update to this branch to provide Windows support.

The output looks like this (Mac OSX sample):

```
Loaded Libraries
  /Users/hhellyer/work/node-sdks/node-v6.5.0-darwin-x64/bin/node
  /System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation
  /usr/lib/libSystem.B.dylib
  /usr/lib/libstdc++.6.dylib
  /usr/lib/libDiagnosticMessagesClient.dylib
  ...
  /usr/lib/libc++abi.dylib
  /usr/lib/libc++.1.dylib
  /Users/hhellyer/work/consumability/node/testscripts/node_modules/nodereport/api.node
```

### DIFF
--- a/src/node_report.cc
+++ b/src/node_report.cc
@@ -27,9 +27,13 @@
 #include <inttypes.h>
 #include <cxxabi.h>
 #include <dlfcn.h>
+#ifdef __linux__
+#include <link.h>
+#endif
 #ifndef _AIX
 #include <execinfo.h>
 #else
+#include <sys/ldr.h>
 #include <sys/procfs.h>
 #endif
 #include <sys/utsname.h>
@@ -46,6 +50,7 @@
 #ifdef __APPLE__
 // Include _NSGetArgv and _NSGetArgc for command line arguments.
 #include <crt_externs.h>
+#include <mach-o/dyld.h>
 #endif
 
 #ifndef _WIN32
@@ -76,6 +81,7 @@ static void PrintResourceUsage(FILE* fp);
 #endif
 static void PrintGCStatistics(FILE* fp, Isolate* isolate);
 static void PrintSystemInformation(FILE* fp, Isolate* isolate);
+static void PrintLoadedLibraries(FILE* fp, Isolate* isolate);
 static void WriteInteger(FILE* fp, size_t value);
 
 // Global variables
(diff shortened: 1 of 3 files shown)


src/node_report.cc
```diff
@@ -1021,6 +1027,65 @@ const static struct {
...
+#elif _AIX
+  const unsigned int buffer_inc = 4096;
+  unsigned int buffer_size = buffer_inc;
+  char* buffer = (char*) malloc(buffer_size);
+  int rc = loadquery(L_GETINFO, buffer, buffer_size);
+  while (rc != 0 && buffer_size < 1024 * 1024 && buffer != nullptr) {
+    free(buffer);
+    buffer_size += buffer_inc;
+    buffer = (char*) malloc(buffer_size);
+    rc = loadquery(L_GETINFO, buffer, buffer_size);
```
richardlau: Do we need to check the return from `malloc()` before calling `loadquery()`?

src/node_report.cc
```diff
@@ -1021,6 +1027,65 @@ const static struct {
...
+  while (name != nullptr) {
+    fprintf(fp, "  %s\n", name);
+    i++;
+    name = _dyld_get_image_name(i);
+  }
+#elif _AIX
+  const unsigned int buffer_inc = 4096;
+  unsigned int buffer_size = buffer_inc;
+  char* buffer = (char*) malloc(buffer_size);
+  int rc = loadquery(L_GETINFO, buffer, buffer_size);
```
richardlau: Do we need to check the return from `malloc()` before calling `loadquery()`?

src/node_report.cc
```diff
@@ -1021,6 +1027,65 @@ const static struct {
...
+      char* member_name = cur_info->ldinfo_filename
+        + strlen(cur_info->ldinfo_filename) + 1;
+      if (*member_name != '\0') {
+        fprintf(fp, "  %s(%s)\n", cur_info->ldinfo_filename, member_name);
+      } else {
+        fprintf(fp, "  %s\n", cur_info->ldinfo_filename);
+      }
+      buf += cur_info->ldinfo_next;
+      cur_info = (ld_info*) buf;
+    } while (cur_info->ldinfo_next != 0);
```
richardlau: I think this will skip over the last entry (i.e. in the previous line cur_info is updated to the last entry, so now `cur_info->ldinfo_next == 0`, but cur_info hasn't yet been processed).

richardlau:
A good sanity check would be to modify the tests to look for `api.node` (the nodereport library) under `Loaded Libraries` in the `System Information` section.

rnchamberlain:
Added commit for Windows support

src/node_report.cc
```diff
@@ -1021,6 +1029,95 @@ const static struct {
...
+                                      FALSE, GetCurrentProcessId());
+  if (process_handle == NULL) {
+    fprintf(fp, "No library information available\n");
+    return;
+  }
+  // Get a list of all the modules in this process
+  DWORD size_1, size_2;
+  // First call to get the size of module array needed
+  if (EnumProcessModules(process_handle, NULL, 0, &size_1)) {
+    HMODULE* modules = (HMODULE*) malloc(size_1);
```
richardlau: Should we be checking the return from malloc prior to using it?

rnchamberlain: Fixed

test/common.js
```diff
@@ -88,6 +88,11 @@ exports.validate = (t, report, options) => {
       t.match(nodeReportSection,
               new RegExp('NodeReport version: ' + nodereportMetadata.version),
               'NodeReport section contains expected NodeReport version');
+      const sysInfoSection = getSection(reportContents, 'System Information');
+      // Find a line which ends with /api.node or \api.node (Unix or Windows paths.)
+      // to see if the library for node report was loaded.
+      t.match(sysInfoSection, /  .*[\/|\\]api\.node/,
```
richardlau: `|` shouldn't be here (unless you meant to allow `|api.node`. Strictly speaking `/` in a character class doesn't need to be escaped so this could be `[/\\]`.

src/node_report.cc
```diff
@@ -1021,6 +1029,106 @@ const static struct {
...
+    if (modules == NULL) {
+      return;  // bail out if malloc failed
+    }
+    // Second call to populate the module array
+    if (EnumProcessModules(process_handle, modules, size_1, &size_2)) {
+      for (int i=0; i < size_1/sizeof(HMODULE) && i < size_2/sizeof(HMODULE); i++) {
```
richardlau: nit: (lack of) spaces around operators.

test/common.js
hhellyer: It should be`(\/|\\)`, we only want to match one kind of path separator. / does need escaping then as it's the delimiter for the regex.

src/node_report.cc
```diff
@@ -1021,6 +1029,106 @@ const static struct {
...
+#elif _AIX
+  // We can't tell in advance how large the buffer needs to be.
+  // Retry until we reach too large a size (1Mb).
+  const unsigned int buffer_inc = 4096;
+  unsigned int buffer_size = buffer_inc;
+  char* buffer = (char*) malloc(buffer_size);
+  int rc = -1;
+  while (buffer != nullptr && rc != 0 && buffer_size < 1024 * 1024) {
+    rc = loadquery(L_GETINFO, buffer, buffer_size);
+    if (rc == 0) {
```
mhdawson: I was going to suggest we check for the error that says the buffer was not big enough specifically but since looking at the API doc its the only error we can expect given the parameters we are passing in its ok as is.

src/node_report.cc
```diff
@@ -1021,6 +1029,106 @@ const static struct {
...
+#elif _WIN32
+  // Windows implementation - get a handle to the process.
+  HANDLE process_handle = OpenProcess(PROCESS_QUERY_INFORMATION | PROCESS_VM_READ,
+                                      FALSE, GetCurrentProcessId());
+  if (process_handle == NULL) {
+    fprintf(fp, "No library information available\n");
+    return;
+  }
+  // Get a list of all the modules in this process
+  DWORD size_1, size_2;
```
mhdawson: Minor nit but I'd suggest we initialize these two values to 0.

src/node_report.cc
mhdawson: agreed it should be 
int i = 0;

src/node_report.cc
rnchamberlain: Done

src/node_report.cc
rnchamberlain: Done

rnchamberlain:
Landed as https://github.com/nodejs/nodereport/commit/082b370a1c7400037724a1723bab18c5fb6e0a50

PR #42, opened 2017-01-20
Fix source directory for install target

richardlau:
Fixes build issues seen on the Node.js CI when trying to build with CITGM: nodejs/citgm#226.

### DIFF
--- a/binding.gyp
+++ b/binding.gyp
@@ -31,7 +31,7 @@
       "copies": [
         {
           "destination": "<(module_root_dir)",
-          "files": ["<(module_root_dir)/build/Release/api.node"]
+          "files": ["<(PRODUCT_DIR)/api.node"]
         }]
     },
   ],


richardlau:
cc @gdams.

gdams:
confirmed to be working in the citgm CI: https://ci.nodejs.org/view/Node.js-citgm/job/gibfahn-citgm-smoker-more-platforms/84/console

rnchamberlain:
The PR builds OK for me on Linux and Windows. Can you explain where PRODUCT_DIR is defined, and why that works for citgm whereas the explicit .../build/Release/... path does not?

richardlau:
To save having to read all of nodejs/citgm#226, here's the failure on the Node.js CI:
```
error:                     | make: Entering directory '/tmp/a27a0182-6fb0-4104-9d68-3cd214c96b02/cf253ebd7215116d02b9159f5ba6df40ddee9507/build'                                                                                                                                                                                                                                               
error:                     | CXX(target) Release/obj.target/api/src/module.o                                                                                                                                                                                                                                                                                                                   
error:                     | CXX(target) Release/obj.target/api/src/node_report.o                                                                                                                                                                                                                                                                                                              
error:                     | make: Leaving directory '/tmp/a27a0182-6fb0-4104-9d68-3cd214c96b02/cf253ebd7215116d02b9159f5ba6df40ddee9507/build'                                                                                                                                                                                                                                                
error:                     |                                                                                                                                                                                                                                                                                                                                                                   
error:                     | make: *** No rule to make target '/tmp/a27a0182-6fb0-4104-9d68-3cd214c96b02/cf253ebd7215116d02b9159f5ba6df40ddee9507/build/Release/api.node', needed by '/tmp/a27a0182-6fb0-4104-9d68-3cd214c96b02/cf253ebd7215116d02b9159f5ba6df40ddee9507/api.node'.  Stop. 
```

Failure was inconsistent on the Node.js CI (sometimes passing, other times failing for a given OS/Linux distribution). Was not able to reproduce locally.

@rnchamberlain `<(PRODUCT_DIR)` is a [predefined variable in gyp](https://gyp.gsrc.io/docs/InputFormatReference.md#Predefined-Variables). Where is `<(module_root_dir)` (the current variable used) defined? I'm not 100% sure why the problem occurs in the first place (as mentioned in the CITGM PR I haven't been able to reproduce on a system I have access to). Perhaps with the help of @gdams/@gibfahn we could capture the generated Makefiles from gyp from the Node.js CI workspaces and compare? 

Alternatively we could remove the `install` target and not copy `api.node` and use something like [`bindings`](https://www.npmjs.com/package/bindings) to load `api.node`.

richardlau:
Also `<(module_root_dir)/build/Release/api.node` wouldn't work for a debug build (which would end up in build/Debug/api.node).

rnchamberlain:
Thanks @richardlau just checking that PRODUCT_DIR was a respectable thing! Works fine for me on Mac + Linux + Win. 

module_root_dir is also pre-defined, but it's a node-gyp thing. It does not seem to be documented, was added here:
https://github.com/nodejs/node-v0.x-archive/commit/1520c7bf5b4586c00ceef95ef9c52adb3ee86217

I did have a look at https://www.npmjs.com/package/bindings a while back but the extra dependency did not seem to bring much extra benefit.

LGTM

rnchamberlain:
Landed as https://github.com/nodejs/nodereport/commit/0c5cc92ca788a43f0d9c3e5fed7f72cf241fa1d7

PR #37, opened 2017-01-12
chore: rename nodereport to node-report

gdams:
This changes nodereport to node-report

Issue #35, opened 2017-01-04
nit? s/nodereport/node-report


joshgav:
I think `node-report` is closer to our typical conventions than `nodereport`, plus it fits nicely with node-inspect, node-heapdump, node-profiler etc. Do you think it would be worth changing the name of the repo, module, etc.? Certainly the sooner the better if we're going to do that.

Personally I think we should change it, I can even put it on my own backlog if others agree.

gibfahn:
I think if we're going to change it the earlier the better.

richardlau:
https://www.npmjs.com/package/node-report Is already taken.

chirag04:
i'm not using `node-report` and seems like nobody else is. i don't even remember if node-report is actually doing anything meaningful.

happy to transfer.

rnchamberlain:
Thanks @chirag04, node-report would be a better name for this project and module

mhdawson:
Provided we can get the name in the appropriate places +1 from me.

chirag04:
I added @rnchamberlain as the owner. Feel to claim it. let me know if i can do anything else to help here.

richardlau:
We're due a semver-major bump for #25 -- Would seem the ideal time to do the rename (if it's not going to take long).

rnchamberlain:
@richardlau agree re https://github.com/nodejs/nodereport/issues/25, but I also have a fix coming up for https://github.com/nodejs/nodereport/issues/6 which changes nodereport behaviour and I'd like to release that at the same time. 

Currently nodereport is at v1.0.7 and node-report is v0.1.0. Would we move node-report to v1.0.0?

gibfahn:
@rnchamberlain It's probably worth creating a github milestone for the next major release.

+1 to `v2.0.0`

richardlau:
#37 handles renaming in the source code. 

Who can rename this repository?

rnchamberlain:
According to https://help.github.com/articles/renaming-a-repository/ someone with the powers will have a gear wheel/settings tab showing on the main repo page https://github.com/nodejs/nodereport. I'm not seeing that myself.

williamkapke:
Looking at the list of Org `owners` that have also been somewhat involved in this group... let's ping @mhdawson and maybe @Fishrock123

rnchamberlain:
Suggested steps required, in sequence:

1. Rename this repo to github.com/nodejs/node-report

[...]

PR #43, opened 2017-01-20
Rename nodereport module to node-report

rnchamberlain:
Provides code, documentation, test and demo changes required for the migration of the nodereport project to node-report. The main changes are in the build, tests and demos that use the module name, and in the name of the report file. Remaining changes are mostly in comments and docs. Internal variable and macro names are not changed.

See issue https://github.com/nodejs/nodereport/issues/35. This PR supercedes PR https://github.com/nodejs/nodereport/pull/37

Builds and tests pass on Linux and Windows. I think we should rename the repo before this lands.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,45 +14,45 @@ Supports Node.js v4, v6 and v7 on Linux, MacOS, Windows and AIX.
 ## Usage
 
 ```bash
-npm install nodereport
-node -r nodereport app.js
+npm install node-report
+node -r node-report app.js
 ```
-A NodeReport will be triggered automatically on unhandled exceptions and fatal
+A report will be triggered automatically on unhandled exceptions and fatal
 error events (for example out of memory errors), and can also be triggered
 by sending a USR2 signal to a Node.js process (Linux/MacOS only).
 
-A NodeReport can also be triggered via an API call from a JavaScript
+A report can also be triggered via an API call from a JavaScript
 application.
 
 ```js
-var nodereport = require('nodereport');
+var nodereport = require('node-report');
 nodereport.triggerReport();
 ```
 The API can be used without adding the automatic exception and fatal error
 hooks and the signal handler, as follows:
 
 ```js
-var nodereport = require('nodereport/api');
+var nodereport = require('node-report/api');
 nodereport.triggerReport();
 ```
 
-Content of the NodeReport consists of a header section containing the event
+Content of the report consists of a header section containing the event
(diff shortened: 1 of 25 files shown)


README.md
```diff
@@ -85,14 +85,14 @@ export NODEREPORT_VERBOSE=yes|no
 
 ## Examples
 
-To see examples of NodeReports generated from these events you can run the
-demonstration applications provided in the nodereport github repository. These are
+To see examples of reports generated from these events you can run the
+demonstration applications provided in the node-report github repository. These are
```
sam-github: you could hyperlink to github

src/module.cc
```diff
@@ -77,13 +77,13 @@ NAN_METHOD(SetEvents) {
   unsigned int previous_events = nodereport_events; // save previous settings
   nodereport_events = ProcessNodeReportEvents(*parameter);
 
-  // If NodeReport newly requested for fatalerror, set up the V8 call-back
+  // If report newly requested for fatalerror, set up the V8 call-back
```
richardlau: Since we're editing -- Should this be "callback" (without the hyphen)?

README.md
rnchamberlain: I have a wider issue, that the same README.md gets used on the npm site and on the github repo, but the information I'd like to have on those two sites is different (the audience is different). So on github I'd like to document the build process, tests and demos, but that's not relevant to the npm user (e.g. they don't have the tests).

Maybe I could put the extra developer info on a page on the github repo wiki, and have a link to that from the README. Or add additional .md files in a docs directory. I wonder what other projects do for this.

src/module.cc
rnchamberlain: Done

README.md
sam-github: most projects put user-focussed info at top of readme, and other info at bottom (and ignore that its not really intended for the npmjs.org audience). I rarely read the info on npmjs.org past the first screen, I usually just jump directly to github even to read using the link at the top-right. You could also rip the dev-focussed stuff into a CONTRIBUTING.md in the top-level, which would get it off npmjs.org. And then link to it from the README, or not? I wouldn't put dev info in a wiki, its too far from the source.

rnchamberlain: OK, I'll add a dev section at the bottom of the readme, under a separate PR, thanks.

rnchamberlain:
This PR got in a mess and would not rebase, closing and raising a new one :-(

PR #45, opened 2017-01-25
Add nodestyle and lint code

### LINKED ISSUE #44: Add lint rules from main nodejs project.
Nodereport should have a lint target developers can run to check the source code formatting rather that having reviewers find the issues.
This would help prevent pull requests from being cluttered up with formatting changes and allow us to focus more on the technical discussions.

Since nodereport (node-report) is a project under Nodejs the rules should be copied over from there so the format is consistent with the main project.

geek:
Closes #44

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,12 +18,15 @@
     "Richard Chamberlain <richard_chamberlain@uk.ibm.com> (https://github.com/rnchamberlain)"
   ],
   "scripts": {
+    "lint": "node-style",
+    "pretest": "npm run lint",
     "test": "tap test/test*.js"
   },
   "bugs": {
     "url": "https://github.com/nodejs/nodereport/issues"
   },
   "devDependencies": {
+    "node-style": "^2.0.0",
     "tap": "^8.0.0"
   }
 }
--- a/.gitignore
+++ b/.gitignore
@@ -42,3 +42,4 @@ NodeReport*.txt
 npm-debug.log
 nodereport-*.tgz
 nodereport_test.log
+.eslintcache
--- a/demo/api_call.js
+++ b/demo/api_call.js
@@ -1,25 +1,28 @@
 // Example - generation of NodeReport via API call
+'use strict';
+
 var nodereport = require('nodereport');
-var http = require("http");
+var http = require('http');
 
 var count = 0;
 
 function my_listener(request, response) {
-  switch(count++) {
-  case 0:
-    response.writeHead(200,{"Content-Type": "text/plain"});
-    response.write("\nRunning NodeReport API demo... refresh page to trigger NodeReport");
-    response.end();
-    break;
-  case 1:
-    response.writeHead(200,{"Content-Type": "text/plain"});
-    // Call the nodereport module to trigger a NodeReport
-    var filename = nodereport.triggerReport();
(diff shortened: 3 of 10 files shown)


hhellyer:
This looks good and I think we should merge it in but I have one major issue and one minor one.
The main issue is this only checks the .js code, on it's own it doesn't close issue #44 as most of the code for this project is the C/C++ which actually generates the report. It's still the right thing to do but it doesn't check the C/C++ code (what I wrote in the description of #44 might not have been clear enough).

Secondly, and more as a proper review comment, is node-style assumed to be globally installed or does it need to be added as a dependency somewhere?

package.json
```diff
@@ -18,12 +18,15 @@
...
   "scripts": {
+    "lint": "node-style",
+    "pretest": "npm run lint",
     "test": "tap test/test*.js"
   },
   "bugs": {
     "url": "https://github.com/nodejs/nodereport/issues"
   },
   "devDependencies": {
+    "node-style": "^1.1.0",
```
gibfahn: @hhellyer node-style is added as a devDependency here (i.e. it won't show up if you `npm i llnode`, but it will show up if you `cd llnode && npm i`.

gibfahn:
@geek I see that you're a maintainer for `node-style`, do you happen to know how different its rules  are from `nodejs/node`'s eslint config? Also what is the benefit to using it over eslint with the defaults (`eslint:recommended`).

geek:
@gibfahn the rules for node-style align with those used for node core. The benefit is consistency and alignment across projects.

mhdawson (approved):
Based on the assertion that these are now the same as used for node core, LGTM

gdams:
I'm struggling to understand why if we want to be consistent with node core why we wouldn't want to use `nodejs/node`'s eslint config

mhdawson (changes requested):
Looking deeper sounds like we need more discussion.  What I'm thinking is that we really just want something which is the same as in node core, if we need something more then we should explain the case for that.  So as George mentioned why would we not just use the nodejs/node's config.

sam-github:
I think #446 is a better approach. The tool being added here may be a nice thing for projects that are not part of `github.com/nodejs/*`, but I don't think we need such a complex dependency when we only need a .eslintrc file.

If nodejs wanted to factor its eslint rules out, eslint already support this easily, https://www.npmjs.com/package/eslint-config-strongloop is an example of what we did for strongloop, it would be easy to do for node's tools, too - though node itself heavily favours "vendoring" its dependencies directly so that the repo is self-contained.

richardlau:
Just tried building this PR on Node.js 4.7.2 and got this:
```
-bash-4.2$ npm test

> nodereport@1.0.7 pretest /home/users/riclau/sandbox/github/nodereport
> npm run lint


> nodereport@1.0.7 lint /home/users/riclau/sandbox/github/nodereport
> node-style

["setTimeout","setInterval"].includes is not a function
```

`Array.includes` isn't supported in V8 4.5. Tracked it down to `node_modules/node-style/node_modules/eslint-plugin-nodejs/lib/timer-arguments.js`.

sam-github:
@richardlau eslint version? We can raise as a bug: https://github.com/eslint/eslint/blob/master/package.json#L120

PR #46, opened 2017-01-25
eslint: add eslint to nodereport

gdams:
Adds eslint to be run with `npm test`. We should use eslintrc to remain consistent with the nodejs community such as nodejs/node and nodejs/citgm. This also allows you to use `npm run lint` or `npm run lint --- --fix` to fix linter errors. 

I have also shortened the timeout to ten seconds for the tap tests

Alternative to: https://github.com/nodejs/nodereport/pull/45

### DIFF
--- a/.eslintrc.yaml
+++ b/.eslintrc.yaml
@@ -1,0 +1,18 @@
+env:
+  node: true
+  es6: true
+
+rules:
+  indent: [2, 2]
+  quotes: [2, single]
+  no-trailing-spaces: 2
+  comma-spacing: [2, { before: false, after: true }]
+  eqeqeq: [2, always]
+  linebreak-style: [2, unix]
+  semi: [2, always]
+  no-console: 0
+  no-inner-declarations: 0
+  no-constant-condition: 0
+  no-regex-spaces: 0
+
+extends: eslint:recommended
--- a/demo/api_call.js
+++ b/demo/api_call.js
@@ -1,21 +1,21 @@
 // Example - generation of NodeReport via API call
 var nodereport = require('nodereport');
-var http = require("http");
+var http = require('http');
 
 var count = 0;
 
 function my_listener(request, response) {
   switch(count++) {
   case 0:
-    response.writeHead(200,{"Content-Type": "text/plain"});
-    response.write("\nRunning NodeReport API demo... refresh page to trigger NodeReport");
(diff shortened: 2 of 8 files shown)


.eslintrc.yaml
```diff
@@ -0,0 +1,18 @@
...
+  comma-spacing: [2, { before: false, after: true }]
+  eqeqeq: [2, always]
+  linebreak-style: [2, unix]
+  semi: [2, always]
+  no-console: 0
+  no-inner-declarations: 0
+  no-constant-condition: 0
+  no-regex-spaces: 0
+
+extends: eslint:recommended
```
sam-github: this isn't the same as nodejs/node: https://github.com/nodejs/node/blob/master/.eslintrc.yaml

Pretty much everyone from Node in a recent thread (that I can't find at the moment, sorry) heavily favours explicit specification, its quite painful updating between eslint versions when using their builtins.

sam-github (changes requested):
Other than the use of `eslint:recommend`, this approach looks simple and very useful.

.eslintrc.yaml
gdams: Most of the sub modules such as citgm don't currently hard code all the rules. The thread was a PR that I submitted the other day. I'm happy to define them all but is it really necessary? If we implement a CI such as travis we would very quickly pick up new rules that break things

sam-github: Isn't the easy thing to just directly copy node's file? I've had horrendous times in the past when eslint bumps majors trying to update, and they used to bump majors or introduce incompatibilities pretty often. From what I saw, I'm not the only one. If its easy, I would just copy the file from nodejs, if that is more complex than it seems (perhaps it relies on some custom rules?), then I've no objection to this. Any eslint is better than none.

sam-github (approved):

.eslintrc.yaml
gdams: Don't hold me to it, but I believe that the community uses several custom rules, i'll double check.

gibfahn: They do, it's in [tools/eslint-rules](https://github.com/nodejs/node/tree/master/tools/eslint-rules).

I don't think it makes a big difference whether we include the rules or not, it shouldn't be that hard to include them if it means possibly less work down the line.

PR #47, opened 2017-01-25
linter: add cpplint

gdams:
This PR adds cpplint to linter all the c/c++ files. It can be run from
`npm test` or directly from `npm run lint`. There are still a few linter errors which I have included below:
```

 ✗ src/module.cc
   3: Include the directory when naming .h files
   16: Using C-style cast.  Use static_cast<int>(...) instead
   266: Using C-style cast.  Use static_cast<int>(...) instead
 ✗ src/node_report.cc
   2: Include the directory when naming .h files
   15: Include the directory when naming .h files
   16: Include the directory when naming .h files
   365: Using C-style cast.  Use reinterpret_cast<char ***>(...) instead
   589: Using C-style cast.  Use reinterpret_cast<LPBYTE *>(...) instead
   1038: Storage class (static, extern, typedef, etc) should be first.
   1096: Using C-style cast.  Use reinterpret_cast<FILE*>(...) instead
   1120: Using C-style cast.  Use reinterpret_cast<char*>(...) instead
   1129: Using C-style cast.  Use reinterpret_cast<char*>(...) instead
   1138: Using C-style cast.  Use reinterpret_cast<ld_info*>(...) instead
   1164: Using C-style cast.  Use reinterpret_cast<HMODULE*>(...) instead
   364: Add #include <string> for string
 ✗ src/node_report.h
   5: Include the directory when naming .h files
   7: Found C system header after other header. Should be: node_report.h, c system, c++ system, other.
   8: Found C system header after other header. Should be: node_report.h, c system, c++ system, other.
   10: Found C system header after other header. Should be: node_report.h, c system, c++ system, other.
```

### DIFF
--- a/cpplint.py
+++ b/cpplint.py
@@ -1,0 +1,6093 @@
+#!/usr/bin/env python
+#
+# Copyright (c) 2009 Google Inc. All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are
+# met:
+#
+#    * Redistributions of source code must retain the above copyright
+# notice, this list of conditions and the following disclaimer.
+#    * Redistributions in binary form must reproduce the above
+# copyright notice, this list of conditions and the following disclaimer
+# in the documentation and/or other materials provided with the
+# distribution.
+#    * Neither the name of Google Inc. nor the names of its
+# contributors may be used to endorse or promote products derived from
+# this software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+# "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+# LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+# A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+# OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+# SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+# LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+# DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+# THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+"""Does google-lint on c++ files.
+
+The goal of this script is to identify places in the code that *may*
+be in non-compliance with google style.  It does not attempt to fix
+up these problems -- the point is to educate.  It does also not
+attempt to find all problems, or to ensure that everything it does
+find is legitimately a problem.
+
+In particular, we can get very confused by /* and // inside strings!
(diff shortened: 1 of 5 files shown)


gdams:
so I have added the cpplint.py to run the python suite that node.js use: https://github.com/nodejs/node/blob/master/tools/cpplint.py

indutny:
I wonder if there is any reason not to use `clang-format` in this project? Linting is good, but fixing issues with one commands is better in my opinion

hhellyer:
Having had a local discussion about this I think the concern was that clang-format wouldn't be available when running citgm builds. (I'd be happy with something that a developer just ran before creating a PR personally but I can see why other people want to integrate it into builds.) 

I think the main concern with this pull request at the moment is cpplint.py is 6153 lines long as it's copied from Node.js it's likely to get out of sync.

gibfahn:
I think ideally we want an npm module we can just have as a devDependency, one that can automatically fix stuff would be great. Not sure if there are any good ones though.

richardlau:
My main objection to `clang-format` is that it isn't part of the standard Node.js addon development environment on platforms such as Windows nor is it (I believe) an installable npm module.

@gdams Where is this version of `cpplint.py` from? It doesn't match the one currently in https://github.com/nodejs/node/blob/master/tools/cpplint.py (e.g. missing https://github.com/nodejs/node/commit/fadf66a02bcbb7cafa480756eeb8df91e1aaaae3).

src/module.cc
```diff
@@ -1,5 +1,6 @@
-#include "node_report.h"
+// Copyright 2017 Nodereport
```
richardlau: Any thoughts on what the copyright attribution should be? @nodejs/post-mortem

cpplint.py
```diff
@@ -0,0 +1,6093 @@
...
+
+    If we have a real absolute path name here we can try to do something smart:
+    detecting the root of the checkout and truncating /path/to/checkout from
+    the name so that we get header guards that don't include things like
+    "C:\Documents and Settings\..." or "/home/username/..." in them and thus
+    people on different computers who have checked the source out to different
+    locations won't see bogus errors.
+    """
+    fullname = self.FullName()
+    # XXX(bnoordhuis) Expects that cpplint.py lives in the tools/ directory.
```
richardlau: Looks like cpplint.py should be in a tools/ directory. Can you move it?

richardlau (commented):
Can you rebase on top of master now that #52 has landed?

src/module.cc
sam-github: @jasnell  you may be bringing all Node.js src copyright statements into alignment with legal requirements, opinions on this?

sam-github:
I've had trouble getting clang-format installed when we used it once before for formatting checking. IIRC, it also had continuously changing formats, so you needed the exact correct version installed. The pain outweighed the gain.

This PR LGTM, except that the copyright header comments don't refer to anyone who can hold copyright... does the linter require them?

package.json
```diff
@@ -18,7 +18,9 @@
     "Richard Chamberlain <richard_chamberlain@uk.ibm.com> (https://github.com/rnchamberlain)"
   ],
   "scripts": {
-    "test": "tap test/test*.js"
+    "test": "npm run lint && npm run tap",
+    "tap": "tap --timeout 10 test/test*.js",
+    "lint": "python cpplint.py **/*.cc **/*.h"
```
sam-github: Sorry, just realized, this should be:

```
   "test": "tap --timeout 10 test/test*.js",
   "pretest": "python cpplint.py **/*.cc **/*.h"
```

And the tap timeout change is unrelated to this PR, shouldn't it be a different PR?

sam-github (changes requested):

src/module.cc
mhdawson: I'll et @jasnell confirm but I think it should be:

```
// Copyright node-report contributors. All rights reserved.
// SPDX-License-Identifier: MIT
```

in line with https://github.com/nodejs/node/pull/10599/files as these are "new" files that were never under the joyent copyright.

gibfahn: I think there was also some discussion about removing the `All rights reserved` from that.

package.json
gibfahn: Why have that as pretest? It means that you can't just do `npm run lint` to lint your code.

Issue #39, opened 2017-01-19
Allowing user to access report content directly rather than reading back from the file.


hhellyer:
We've had a feature request to allow users of the nodereport api to get the content of the nodereport file directly rather than having to read it back in from the file. This would make it easier to embed nodereport in other tools.

I'm going to take a look at this but up front I can think of some important limitations:
 - This will only work on API calls. We can't send the content of nodereport to a  callback on a fatal error. (At that point Node.js isn't available for use.)
 
 - We will have to return all the content in one go, synchronously. If we stream it  back bit by bit to a callback then code is running and changing the state of Node.js while  we try and report it's state. The response from nodereport is a snapshot of a point in time.

 I'm going to do some more investigation but any other feedback into the design would be welcome.

rnchamberlain:
Maybe a separate issue, but some work is needed on the error handling for the API use-case. If a nodereport is triggered via the API, we should probably use Nan::ThrowError() rather than writing messages to stderr.

mhdawson:
Agreed that we can't have the state changing as we generate the report so it does need to be a one shot generation.    

Assuming we have to generate in full, then we'll need a string which is the full size of the nodereport.  However having that we could still provide a streaming api that returns the content of that string, but that might just lead to needing even more memory.

May just make sense to have a method that returns the full node report and then a second one that is called to free the corresponding buffer.

rnchamberlain:
Delivered in npm https://www.npmjs.com/package/node-report v2.1.0

hhellyer:
This has been released in v2.1.0 - closing issue.

PR #48, opened 2017-01-30
Provide getReport API to return the contents of node-report.

hhellyer:
These changes add a new function to the node-report API - getReport() to address issue #39. The getReport() function returns the string contents of node report to the caller instead of writing it to a file and returning the file name. This makes it easier for other modules to build on node-report.

This is limited to non-fatal uses of node-report, there is no event listener API. A Node.js program can’t listen on a fatal event callback and get the contents of a node-report as at that point Node.js is broken. Similarly the output cannot be streamed, it is returned as one object representing a snapshot of a point in time, as processing the data via a callback while it was still being gathered would change the state of the process we are reporting on.

To implement this I switched from using fprintf for output to C++ streams. This makes it trivial to switch between using a file stream or a string stream for output and the getReport path just passing a different stream object to the triggerReport path. There were two places in node-report where we passed our file stream to internal functions for them to write to.

- In one case in the stack walking code there is no other suitable API so we use a temporary file and read it back in.
- In another location where were were walking the libuv event loop via uv_print_all_handles. I have switched API’s to uv_walk, which allows us to look directly at each handle.
This loses the ability to see internal handles, however there only appear to be a couple of those and they are probably uninteresting. We gain the ability to get hold of the real uv_handle_t structures on the event loop and inspect them further which we can exploit in a future enhancement to give the user information about things like tcp connections and write queue sizes. (Anything publicly available via the libuv public API’s.) That would probably have been of sufficient benefit to switch API’s anyway.

This change also fixes a bug where when node-report always closed the handle it was writing to, even if it was using stderr/stdout.

### DIFF
--- a/src/module.cc
+++ b/src/module.cc
@@ -78,7 +78,7 @@ NAN_METHOD(GetReport) {
   std::ostringstream out;
 
   GetNodeReport(isolate, kJavaScript, "JavaScript API", __func__, out);
-  // Return value is the contents of a NodeReport as a string.
+  // Return value is the contents of a report as a string.
   info.GetReturnValue().Set(Nan::New(out.str()).ToLocalChecked());
 }
 
--- a/src/module.cc
+++ b/src/module.cc
@@ -1,5 +1,6 @@
 #include "node_report.h"
 #include <nan.h>
+#include <sstream>
 
 namespace nodereport {
 
@@ -65,6 +66,20 @@ NAN_METHOD(TriggerReport) {
     // Return value is the report filename
     info.GetReturnValue().Set(Nan::New(filename).ToLocalChecked());
   }
+}
+
+/*******************************************************************************
+ * External JavaScript API for returning a report
+ *
+ ******************************************************************************/
+NAN_METHOD(GetReport) {
+  Nan::HandleScope scope;
+  v8::Isolate* isolate = info.GetIsolate();
+  std::ostringstream out;
+
+  GetNodeReport(isolate, kJavaScript, "JavaScript API", __func__, out);
+  // Return value is the contents of a NodeReport as a string.
+  info.GetReturnValue().Set(Nan::New(out.str()).ToLocalChecked());
 }
 
 /*******************************************************************************
@@ -392,6 +407,8 @@ void Initialize(v8::Local<v8::Object> exports) {
 
   exports->Set(Nan::New("triggerReport").ToLocalChecked(),
                Nan::New<v8::FunctionTemplate>(TriggerReport)->GetFunction());
+  exports->Set(Nan::New("getReport").ToLocalChecked(),
+               Nan::New<v8::FunctionTemplate>(GetReport)->GetFunction());
   exports->Set(Nan::New("setEvents").ToLocalChecked(),
                Nan::New<v8::FunctionTemplate>(SetEvents)->GetFunction());
   exports->Set(Nan::New("setSignal").ToLocalChecked(),
--- a/src/node_report.cc
+++ b/src/node_report.cc
@@ -459,7 +459,7 @@ void TriggerNodeReport(Isolate* isolate, DumpEvent event, const char* message, c
 
   std::cerr << "Node.js report completed\n";
   if (name != nullptr) {
-    snprintf(name, NR_MAXNAME + 1, "%s", filename);  // return the NodeReport file name
+    snprintf(name, NR_MAXNAME + 1, "%s", filename);  // return the report file name
   }
 
 }
@@ -525,7 +525,7 @@ static void WriteNodeReport(Isolate* isolate, DumpEvent event, const char* messa
   pid_t pid = getpid();
 #endif
 
-  // File stream opened OK, now start printing the NodeReport content, starting with the title
+  // File stream opened OK, now start printing the report content, starting with the title
   // and header information (event, filename, timestamp and pid)
   out << "================================================================================\n";
   out << "==== Node Report ===============================================================\n";
--- a/src/node_report.cc
+++ b/src/node_report.cc
@@ -7,6 +7,8 @@
 #include <string.h>
 #include <stdio.h>
 #include <stdlib.h>
+#include <iostream>
+#include <fstream>
 
 #if !defined(_MSC_VER)
 #include <strings.h>
@@ -69,20 +71,27 @@ using v8::Message;
 using v8::String;
 using v8::V8;
 
-// Internal/static function declarations
-static void PrintCommandLine(FILE* fp);
-static void PrintVersionInformation(FILE* fp, Isolate* isolate);
-static void PrintJavaScriptStack(FILE* fp, Isolate* isolate, DumpEvent event, const char* location);
-static void PrintStackFromStackTrace(FILE* fp, Isolate* isolate, DumpEvent event);
-static void PrintStackFrame(FILE* fp, Isolate* isolate, Local<StackFrame> frame, int index, void* pc);
-static void PrintNativeStack(FILE* fp);
-#ifndef _WIN32
-static void PrintResourceUsage(FILE* fp);
+#ifdef _WIN32
+typedef SYSTEMTIME TIME_TYPE;
+#else  // UNIX, OSX
+typedef struct tm TIME_TYPE;
 #endif
-static void PrintGCStatistics(FILE* fp, Isolate* isolate);
-static void PrintSystemInformation(FILE* fp, Isolate* isolate);
-static void PrintLoadedLibraries(FILE* fp, Isolate* isolate);
-static void WriteInteger(FILE* fp, size_t value);
+
+// Internal/static function declarations
+static void WriteNodeReport(Isolate* isolate, DumpEvent event, const char* message, const char* location, char* filename, std::ostream &out, TIME_TYPE* time);
+static void PrintCommandLine(std::ostream& out);
+static void PrintVersionInformation(std::ostream& out);
+static void PrintJavaScriptStack(std::ostream& out, Isolate* isolate, DumpEvent event, const char* location);
+static void PrintStackFromStackTrace(std::ostream& out, Isolate* isolate, DumpEvent event);
+static void PrintStackFrame(std::ostream& out, Isolate* isolate, Local<StackFrame> frame, int index, void* pc);
+static void PrintNativeStack(std::ostream& out);
+#ifndef _WIN32
+static void PrintResourceUsage(std::ostream& out);
+#endif
+static void PrintGCStatistics(std::ostream& out, Isolate* isolate);
+static void PrintSystemInformation(std::ostream& out, Isolate* isolate);
+static void PrintLoadedLibraries(std::ostream& out, Isolate* isolate);
+static void WriteInteger(std::ostream& out, size_t value);
 
 // Global variables
 static int seq = 0;  // sequence number for report filenames
@@ -92,11 +101,7 @@ static char report_filename[NR_MAXNAME + 1] = "";
 static char report_directory[NR_MAXPATH + 1] = ""; // defaults to current working directory
 std::string version_string = UNKNOWN_NODEVERSION_STRING;
 std::string commandline_string = "";
-#ifdef _WIN32
-static SYSTEMTIME loadtime_tm_struct; // module load time
-#else  // UNIX, OSX
-static struct tm loadtime_tm_struct; // module load time
-#endif
+static TIME_TYPE loadtime_tm_struct; // module load time
 
 /*******************************************************************************
  * Functions to process node-report configuration options:
@@ -125,7 +130,7 @@ unsigned int ProcessNodeReportEvents(const char* args) {
       event_flags |= NR_APICALL;
       cursor += sizeof("apicall") - 1;
     } else {
-      fprintf(stderr, "Unrecognised argument for node-report events option: %s\n", cursor);
+      std::cerr << "Unrecognised argument for node-report events option: " << cursor << "\n";
       return 0;
     }
     if (*cursor == '+') {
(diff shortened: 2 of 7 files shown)


src/node_report.cc
```diff
@@ -598,35 +677,35 @@ static void PrintVersionInformation(FILE* fp, Isolate* isolate) {
...
           os_name = (isServer ? "Windows Server" : "Windows Client");
       }
-      fprintf(fp, "\nOS version: %s\n", os_name);
+      out <<  "\nOS version: " << os_name << "\n";
 
       if (os_info->sv101_comment != NULL) {
-        fprintf(fp, "\nMachine: %ls %ls\n", os_info->sv101_name,
-                os_info->sv101_comment);
+        out << "\nMachine: " << os_info->sv101_name << " " <<  os_info->sv101_comment << "\n";
+
```
mhdawson: extra blank line, remove

src/node_report.cc
```diff
@@ -635,35 +714,47 @@ static void PrintVersionInformation(FILE* fp, Isolate* isolate) {
...
+  case kJavaScript: {
     // Print the stack using Message::PrintCurrentStackTrace() API
-    Message::PrintCurrentStackTrace(isolate, fp);
+    std::FILE *stack_fp = std::tmpfile();
+    if( stack_fp != nullptr ) {
+      char stack_buf[64];
+      Message::PrintCurrentStackTrace(isolate, stack_fp);
+      std::fflush(stack_fp);
+      std::rewind(stack_fp);
+      while( std::fgets(stack_buf, sizeof(stack_buf), stack_fp) != nullptr ) {
```
mhdawson: extra spaces after while( and before closing )

src/node_report.cc
```diff
@@ -1132,12 +1247,13 @@ static void WriteInteger(FILE* fp, size_t value) {
...
-      fprintf(fp, "%u", thousandsStack[i]);
+      out << thousandsStack[i];
     } else {
-      fprintf(fp, "%03u", thousandsStack[i]);
+      snprintf(buf, sizeof(buf), "%03u", thousandsStack[i]);
+      out << buf;
     }
     if (i > 0) {
-       fprintf(fp, ",");
+       out << ",";
```
mhdawson: I think there is an extra indent space

src/node_report.cc
```diff
@@ -430,126 +435,200 @@ void TriggerNodeReport(Isolate* isolate, DumpEvent event, const char* message, c
...
 
   // Print Node.js and deps component versions
-  fprintf(fp, "\n%s", version_string.c_str());
+  out << "\n" << version_string;
 
   // Print NodeReport version
   // e.g. NodeReport version: 1.0.6 (built against Node.js v6.9.1)
-  fprintf(fp, "\nNodeReport version: %s (built against Node.js v%s)\n",
-          NODEREPORT_VERSION, NODE_VERSION_STRING);
+  out << "\nNodeReport version: " << NODEREPORT_VERSION << " (built against Node.js v" << NODE_VERSION_STRING << ")\n";
```
mhdawson: I'm guessing this line is now too long so we need some wrapping

src/node_report.cc
```diff
@@ -712,36 +803,39 @@ static void PrintStackFromStackTrace(FILE* fp, Isolate* isolate, DumpEvent event
...
     return;
   }
 
   if (fn_name_s.length() == 0) {
-    fprintf(fp, "%s:%i:%i\n", *script_name, line_number, column);
+    out  << *script_name << ":" << line_number << ":" << column << "\n";
   } else {
     if (frame->IsConstructor()) {
-      fprintf(fp, "%s [constructor] (%s:%i:%i)\n", *fn_name_s, *script_name, line_number, column);
+      out  << *fn_name_s << " [constructor] (" << *script_name << ":" << line_number << ":" << column << ")\n";
```
mhdawson: Another long line, overall I think we need to go through an make sure we've not pushed lines over the limit, especially since we may turn on the linter soon.

src/node_report.cc
```diff
@@ -635,35 +714,47 @@ static void PrintVersionInformation(FILE* fp, Isolate* isolate) {
...
     // Print the stack using Message::PrintCurrentStackTrace() API
-    Message::PrintCurrentStackTrace(isolate, fp);
+    std::FILE *stack_fp = std::tmpfile();
+    if( stack_fp != nullptr ) {
+      char stack_buf[64];
+      Message::PrintCurrentStackTrace(isolate, stack_fp);
+      std::fflush(stack_fp);
+      std::rewind(stack_fp);
+      while( std::fgets(stack_buf, sizeof(stack_buf), stack_fp) != nullptr ) {
+        out << stack_buf;
```
mhdawson: It feels a bit painful to do this even in the case of when we are writing out to a file as it gets written to a file twice.  Having said that I don't have a good idea for how to avoid and be able to easily switch the output of the stream.

This might be something to put on our todo list in terms of working with V8 to get an API that can be used without having to write to a file.

hhellyer:
The format of the output for the handles section (and everything else) should be the same as before. The handles section has lost one flags column but that indicated if a flag was internal to libuv and we don't get those anymore so it would never bet set.

I kept the format the same as I was trying to do a simple like for like swap from fprintf to streams. For the record though I don't believe we are keeping the format tightly locked down - it's intended to be human readable and anyone parsing it with other scripts does so at their own peril. :-)

I agree that using a temporary file is messy, it's also a bit awkward that depending on the type of failure we use two different methods of obtaining a stack. Hopefully having a use case like node-report will let us push some improvements back into Node itself.

Regarding the code format changes, I'll fix anything this introduces but I tried various formatters on nodereport.cc and it's going to need a fair few changes when that's turned on anyway. An 80 character limit is going to mess up all the print statements that print 80 `'='` characters as dividers. (We can fix that by having a constant or doing `std::string(80, '=')` and probably several other ways but that's a separate PR.)

I'll push a tidy up to the formatting issues a bit later on.

src/node_report.cc
```diff
@@ -635,35 +716,47 @@ static void PrintVersionInformation(FILE* fp, Isolate* isolate) {
...
+      char stack_buf[64];
+      Message::PrintCurrentStackTrace(isolate, stack_fp);
+      std::fflush(stack_fp);
+      std::rewind(stack_fp);
+      while (std::fgets(stack_buf, sizeof(stack_buf), stack_fp) != nullptr) {
+        out << stack_buf;
+      }
+      // Calling close on a file from tmpfile *should* delete it.
+      std::fclose(stack_fp);
+    }
```
richardlau: I suppose there's no fallback here if `std::tmpfile()` fails? Should we emit some sort of message, e.g. "Unable to obtain stack trace"?

richardlau:
Also would be good to have tests and documentation.

src/node_report.cc
```diff
@@ -430,126 +435,201 @@ void TriggerNodeReport(Isolate* isolate, DumpEvent event, const char* message, c
...
+}
+
+static void walkHandle(uv_handle_t* h, void* arg) {
+  std::string type;
+  std::string data = "";
+  std::ostream* out = reinterpret_cast<std::ostream*>(arg);
+  char buf[64];
+
+  // List all the types so we get a compile warning if we've missed one,
+  // (using default: supresses the compiler warning.)
```
hhellyer: I've listed all the types so I can ommit the default: statement. If a new type is added the compiler will warn it's not being handled rather than just let the default: swallow it.

hhellyer:
I've pushed an update with tests and doc changed. I've ran it via citgm and verified the new test is running there too:
https://ci.nodejs.org/view/post-mortem/job/nodereport-continuous-integration/41/
( @mhdawson - citgm seems to be working nicely here.)

hhellyer:
(Test file changes look better if we ignore whitespace: https://github.com/nodejs/node-report/pull/48/files?w=1 )

src/module.cc
```diff
@@ -68,6 +69,21 @@ NAN_METHOD(TriggerReport) {
...
+ * External JavaScript API for returning a NodeReport
+ *
+ ******************************************************************************/
+NAN_METHOD(GetReport) {
+  Nan::HandleScope scope;
+  v8::Isolate* isolate = info.GetIsolate();
+  std::ostringstream out;
+
+  GetNodeReport(isolate, kJavaScript, "JavaScript API", __func__, out);
+  // Return value is the NodeReport filename
```
rnchamberlain: This comment looks wrong, probably cut&pasted, the return value for this API is the report string

src/node_report.cc
```diff
@@ -752,14 +853,15 @@ static void PrintStackFrame(FILE* fp, Isolate* isolate, Local<StackFrame> frame,
...
   void* frames[64];
-  fprintf(fp, "\n================================================================================");
-  fprintf(fp, "\n==== Native Stack Trace ========================================================\n\n");
+  out << "\n================================================================================";
+  out << "\n==== Native Stack Trace ========================================================\n\n";
 
   HANDLE hProcess = GetCurrentProcess();
   SymSetOptions(SYMOPT_LOAD_LINES | SYMOPT_UNDNAME | SYMOPT_DEFERRED_LOADS);
   SymInitialize(hProcess, nullptr, TRUE);
+  char buf[64];
```
rnchamberlain: I wonder if 64 chars will always be enough for symbol+filename etc. Maybe we can afford to be more generous. The snprintf() will just truncate, so it's safe anyway.

src/node_report.cc
```diff
@@ -795,52 +900,54 @@ void PrintNativeStack(FILE* fp) {
...
 /*******************************************************************************
  * Function to print a native stack backtrace - Linux/OSX
  *
  ******************************************************************************/
-void PrintNativeStack(FILE* fp) {
+void PrintNativeStack(std::ostream& out) {
   void* frames[256];
-  fprintf(fp, "\n================================================================================");
-  fprintf(fp, "\n==== Native Stack Trace ========================================================\n\n");
+  char buf[64];
```
rnchamberlain: ditto

src/module.cc
hhellyer: Good spot, I'll fix that.

src/node_report.cc
hhellyer: That was a bug, `buf` was being filled and the output was being truncated on Windows.
I've re-written the block so only the `"%2d: [pc=0x%p]"` part is formatted to `buf` as that's the only part that required the format specifiers, the rest can be written with `<<`.

src/node_report.cc
hhellyer: On this path the buffer only ever has to hold the result of `"%2d: [pc=0x%p]"` which should fit into 64 characters.

src/node_report.cc
```diff
@@ -878,16 +878,12 @@ void PrintNativeStack(std::ostream& out) {
       DWORD dwOffset = 0;
       IMAGEHLP_LINE64 line;
       line.SizeOfStruct = sizeof(line);
+      snprintf(buf, sizeof(buf), "%2d: [pc=0x%p]", i, reinterpret_cast<void*>(pSymbol->Address));
+      out << buf << " " << pSymbol->Name << " [+";
```
richardlau: I wonder if it's worth inserting a comment with sample output lines (one for each branch of the nested `if` statement) as it's now harder to tell what the output is supposed to look like as the formatting is split between the `snprintf` prior to the if statement and `<<` statements both here and in the nested `if` statement.

src/module.cc
```diff
@@ -78,7 +78,7 @@ NAN_METHOD(GetReport) {
   std::ostringstream out;
 
   GetNodeReport(isolate, kJavaScript, "JavaScript API", __func__, out);
-  // Return value is the NodeReport filename
+  // Return value is the contents of a NodeReport as a string.
```
richardlau: Suggestion: Since this is a newly introduced line, replace "NodeReport" with "report" so it doesn't need to change again for #52.

richardlau:
I'm :+1: to using C++ streams, but will ask the following question so that we've documented that we've considered it:

Since we have to write out a temporary file anyway and read it back in to get the stack traces, presumably there would have been much less code change if we were to simply write out the whole report to a temporary file and then read the report back in and return the contents?

richardlau:
Also I don't think this has been noted previously but a nice side effect of this PR is that we now get a libuv handle summary on Windows. :tada:

hhellyer:
Writing to a file and then reading back in is (frankly) a bit rubbish. It's something we are stuck with for the moment and will hopefully resolve. It's only used in one of the paths where we write a stack trace. (Having to use different methods depending on the state of v8 is also annoying.)

With the changes in this PR there are plenty of situations where we avoid writing a temporary file at all. Once this is merged I'm happy to raise an issue we can investigate removing (or improving) the call. v8.h already has a TODO to fix it:
```
  // TODO(1245381): Print to a string instead of on a FILE.
  static void PrintCurrentStackTrace(Isolate* isolate, FILE* out);
```

src/module.cc
hhellyer: If #52 is merged first I'll need to make changes here, if this one is merged first then the renaming one will need updates. I'll make sure it gets done at that point with everything else rather than do just one of the (many) changes I'll probably need to make ahead of time.

richardlau: #52 has now been merged, please can you rebase and update accordingly?

hhellyer:
I had done the rebase (or just flattened it onto the new master) but I missed a couple of occurrences of "NodeReport". I've fixed those.

rnchamberlain:
Landed as https://github.com/nodejs/node-report/commit/c3ee5a011cb77ff9d3b34846fd1469a8e4ca0ebb

rnchamberlain:
CI build/test, on Node v6: https://ci.nodejs.org/view/post-mortem/job/nodereport-continuous-integration/49/

PR #53, opened 2017-02-06
makefile: add makefile to linter c++ code

gdams:
This can be run using `make linter` but will also be executed by
default with the `npm test` command

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,13 @@
+PYTHON ?= python
+
+all: lint
+
+lint: cpplint
+
+cpplint.py:
+	curl -o cpplint.py "https://raw.githubusercontent.com/nodejs/node/master/tools/cpplint.py"
+
+cpplint: cpplint.py
+	python cpplint.py **/*.cc **/*.h
+
+.PHONY: all cpplint lint r
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "Richard Chamberlain <richard_chamberlain@uk.ibm.com> (https://github.com/rnchamberlain)"
   ],
   "scripts": {
-    "test": "tap test/test*.js"
+    "test": "make lint && tap test/test*.js"
   },
   "bugs": {
     "url": "https://github.com/nodejs/nodereport/issues"
--- a/.gitignore
+++ b/.gitignore
@@ -33,6 +33,9 @@
 *.project
 *.cproject
 
+# Linter files
+cpplint.py
+
 # Build directories
 /build/
 /node_modules/


Makefile
```diff
@@ -0,0 +1,9 @@
+PYTHON ?= python
+
+all: linter
+
+cpplint.py:
+	curl -o cpplint.py "https://raw.githubusercontent.com/nodejs/node/master/tools/cpplint.py"
```
bnoordhuis: You may want to pin it to a specific release (e.g. `v7.5.0` instead of `master`) in case it gets updated to different settings.

Makefile
```diff
@@ -0,0 +1,9 @@
...
+
+linter: cpplint.py
```
bnoordhuis: In nodejs/node, the target is called `cpplint` and there's an overarching `lint` target for linting C++ and JS code; you may want to follow that example here.

Can you add a `.PHONY: all cpplint lint` rule?

Makefile
```diff
@@ -0,0 +1,13 @@
...
+
+lint: cpplint
+
+cpplint.py:
+	curl -o cpplint.py "https://raw.githubusercontent.com/nodejs/node/master/tools/cpplint.py"
+
+cpplint: cpplint.py
+	python cpplint.py **/*.cc **/*.h
+
+.PHONY: all cpplint lint r
```
richardlau: `r`?

package.json
```diff
@@ -18,7 +18,7 @@
     "Richard Chamberlain <richard_chamberlain@uk.ibm.com> (https://github.com/rnchamberlain)"
   ],
   "scripts": {
-    "test": "tap test/test*.js"
+    "test": "make lint && tap test/test*.js"
```
richardlau: Requiring `make` breaks `npm test` on Windows:

```
C:\work\node\github\nodereport>npm test

> nodereport@1.0.7 test C:\work\node\github\nodereport
> make lint && tap test/test*.js

'make' is not recognized as an internal or external command,
operable program or batch file.

C:\work\node\github\nodereport>
```

I would prefer, if possible, that we are able to run the linter on all supported platforms. It seems from https://github.com/nodejs/node/issues/5514#issuecomment-191188708 that in the main Node.js project `cpplint` is not currently run on Windows, but it sounds more like a tooling problem that an issue with `cpplint.py` itself. Prior to this PR we don't have explicit `makefile`/`vcbuild.bat` since we rely on `binding.gyp` and `node-gyp` -- Perhaps downloading and running cpplint could be a script written in Node.js?

gibfahn: Agreed, I don't see a reason to add a `Makefile`, it's another entry point to the module, which seems unnecessary as we already have a `package.json`.

mhdawson: I agree if we can use a Node.js script instead of make that would simply supporting it across platforms.

Issue #55, opened 2017-02-07
Error building nodereport


curtisr7:
While running `npm install` against appmetrics-dash I'm getting an error [1] while building nodereport. `appmetrics` depends on https://github.com/hhellyer/nodereport.git#stream_output_dev

````
$ uname -a
Darwin rickmbp 16.3.0 Darwin Kernel Version 16.3.0: Thu Nov 17 20:23:58 PST 2016; root:xnu-3789.31.2~1/RELEASE_X86_64 x86_64
````

[1]
````
$ npm install
-
> nodereport@1.0.7 install /code/appmetrics-dash/node_modules/nodereport
> node-gyp rebuild

  CXX(target) Release/obj.target/api/src/node_report.o
clang: error: invalid deployment target for -stdlib=libc++ (requires OS X 10.7 or later)
make: *** [Release/obj.target/api/src/node_report.o] Error 1
gyp ERR! build error 
gyp ERR! stack Error: `make` failed with exit code: 2
gyp ERR! stack     at ChildProcess.onExit (/usr/local/lib/node_modules/npm/node_modules/node-gyp/lib/build.js:270:23)
gyp ERR! stack     at emitTwo (events.js:87:13)
gyp ERR! stack     at ChildProcess.emit (events.js:172:7)
gyp ERR! stack     at Process.ChildProcess._handle.onexit (internal/child_process.js:200:12)
gyp ERR! System Darwin 16.3.0
gyp ERR! command "/usr/local/Cellar/node/4.1.1/bin/node" "/usr/local/lib/node_modules/npm/node_modules/node-gyp/bin/node-gyp.js" "rebuild"
gyp ERR! cwd /code/appmetrics-dash/node_modules/nodereport
gyp ERR! node -v v4.1.1
gyp ERR! node-gyp -v v3.0.1
gyp ERR! not ok 
npm WARN optional dep failed, continuing nodereport@1.0.7
````

richardlau:
This is probably because the deployment target is set as 10.5 in Node.js v4.x: https://github.com/nodejs/node/blob/v4.x/common.gypi#L355

richardlau:
Reproduces on the CI with Node.js v4: https://ci.nodejs.org/view/post-mortem/job/nodereport-continuous-integration/45/MACHINE=osx1010/console

hhellyer:
Ok, so the problem seems to be the change from https://github.com/nodejs/node-report/pull/31/files
[...]

PR #56, opened 2017-02-08
mac: Fix compilation errors against Node > v5

hhellyer:
Only use libc++ on Mac when Node itself supports it. (Node 6 and higher.)
Don’t use std::prev as that requires libc++

This should resolve issue #55.

CI runs on Node v4:
https://ci.nodejs.org/view/post-mortem/job/nodereport-continuous-integration/46/
and Node v6:
https://ci.nodejs.org/view/post-mortem/job/nodereport-continuous-integration/47/
OSX builds passed for both of those.

### DIFF
--- a/binding.gyp
+++ b/binding.gyp
@@ -1,4 +1,7 @@
 {
+  'variables': {
+        'node_major_version': '<!(node -e "console.log(process.version.substr(1,1))")',
+  },
   "targets": [
     {
       "target_name": "api",
@@ -13,7 +16,7 @@
           "libraries": [ "dbghelp.lib", "Netapi32.lib", "PsApi.lib" ],
           "dll_files": [ "dbghelp.dll", "Netapi32.dll", "PsApi.dll" ],
         }],
-        ["OS=='mac'", {
+        ["OS=='mac' and node_major_version > 5", {
           "xcode_settings": {
             'CLANG_CXX_LIBRARY': 'libc++',
             'CLANG_CXX_LANGUAGE_STANDARD':'c++11',
--- a/binding.gyp
+++ b/binding.gyp
@@ -1,7 +1,4 @@
 {
-  'variables': {
-        'node_major_version': '<!(node -e "console.log(process.version.substr(1,1))")',
-  },
   "targets": [
     {
       "target_name": "api",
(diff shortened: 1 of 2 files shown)


binding.gyp
```diff
@@ -1,4 +1,7 @@
 {
+  'variables': {
+        'node_major_version': '<!(node -e "console.log(process.version.substr(1,1))")',
```
richardlau: We can shorten this line by using `-p` instead of `-e`:
```
       'node_major_version': '<!(node -p "process.version.substr(1,1)")',
```

src/node_report.cc
```diff
@@ -264,26 +264,21 @@ void SetVersionString(Isolate* isolate) {
...
-    version_string += component_version;
-    line_length += length;
-    if (it != *std::prev(comp_versions.end())) {
-      version_string += ", ";
-      line_length += 2;
+    std::string comp_version_string = it.first;
+    comp_version_string += ": ";
+    comp_version_string += it.second;
+    versions += separator;
+    if (wrap - (versions.length() % wrap) < comp_version_string.length()) {
```
richardlau: The separator is unconditionally appended before the wrapping check, so could take the line over the wrap if the line was right up against the wrap prior to the separator being appended.

hhellyer: I don't think it matters, the line stays at a sensible length. We don't worry about wrapping elsewhere in the file and plenty of other lines go over 80 characters (OS version and stack trace at least) and we don't wrap those. (Wrapping the stack trace would be horrible.)

Possibly we should just remove the wrapping code for consistency or give each version it's own line.

richardlau: I had one component/version per line originally in https://github.com/nodejs/node-report/pull/30 but was talked out of it.

hhellyer:
I'll push the change to use -p instead of -e. Shall I change the version check to "node_major_version > 6" as well so we do match the setting for Node.js?

We can live with the deprecation warning if we understand why we are seeing it.

richardlau:
If we check "node_major_version > 6" then we might as well revert https://github.com/nodejs/node-report/pull/31 entirely and just pick up whatever settings we inherit from Node.js' `common.gypi`.

hhellyer:
Remove the binding.gyp change sounds right though I'm a little unclear then as to why we build on node 6 at all without that change but fail on 4 if Node.js didn't change what we inherit until 7.

We still need to remove std::prev as that doesn't seem to be supported at all on the older C++ library.

richardlau:
The deployment target was bumped in Node.js 6 to 10.7 (https://github.com/nodejs/node/pull/6402) but didn't switch to libc++ (that change came with 7).

hhellyer:
Ok, done. The change to master is now to remove the xcode_settings from binding.gyp and remove the use of std::prev.
I've tested against node v4, v6 and v7 -  it works on all of them but only v4 and v6 give the clang deprecation warning. (The fix for that is for users to upgrade to node v7!)

rnchamberlain:
Landed as https://github.com/nodejs/node-report/commit/cd22ccfcd44540fd9dc7cbc8940cf21f84d13c50

PR #57, opened 2017-02-09
aix: improve readability of os version

richardlau:
uname on AIX reports, e.g. 6.1 as version == 6, release == 1. 
The code was previously printing this as:
```
OS version: AIX 1 6
```
Fix so that on AIX this is now:
```
OS version: AIX 6.1
```

https://github.com/nodejs/node/pull/10245 fixed a similar area in Node.js.

### DIFF
--- a/test/test-os-version.js
+++ b/test/test-os-version.js
@@ -1,0 +1,27 @@
+'use strict';
+
+// Testcase for validating reported OS version
+
+const common = require('./common.js');
+const nodereport = require('../');
+const os = require('os');
+const tap = require('tap');
+
+const os_map = {
+  'aix': 'AIX',
+  'darwin': 'Darwin',
+  'linux': 'Linux',
+  'win32': 'Windows',
+};
+const os_name = os_map[os.platform()];
+const report_str = nodereport.getReport();
+const version_str = report_str.match(/OS version: .*(?:\r*\n)/);
+if (common.isWindows()) {
+  tap.match(version_str, new RegExp('OS version: ' + os_name), 'Checking OS version');
+} else if (common.isAIX() && !os.release().includes('.')) {
+  // For Node.js prior to os.release() fix for AIX:  https://github.com/nodejs/node/pull/10245
+  tap.match(version_str, new RegExp('OS version: ' + os_name + ' \\d+.' + os.release()), 'Checking OS version');
+} else if (!common.isWindows()) {
+  tap.match(version_str, new RegExp('OS version: ' + os_name + ' .*' + os.release()), 'Checking OS version');
+}
+
--- a/test/test-os-version.js
+++ b/test/test-os-version.js
@@ -17,11 +17,17 @@ const os_name = os_map[os.platform()];
 const report_str = nodereport.getReport();
 const version_str = report_str.match(/OS version: .*(?:\r*\n)/);
 if (common.isWindows()) {
-  tap.match(version_str, new RegExp('OS version: ' + os_name), 'Checking OS version');
+  tap.match(version_str,
+            new RegExp('OS version: ' + os_name), 'Checking OS version');
 } else if (common.isAIX() && !os.release().includes('.')) {
-  // For Node.js prior to os.release() fix for AIX:  https://github.com/nodejs/node/pull/10245
-  tap.match(version_str, new RegExp('OS version: ' + os_name + ' \\d+.' + os.release()), 'Checking OS version');
-} else if (!common.isWindows()) {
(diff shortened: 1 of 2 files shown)


richardlau:
I'll add a test, just need to think of a sensible thing to check (maybe for AIX that it conforms the regex `/AIX\ \d+\.\d+/`). Can't rely on Node.js to get the expected string (https://github.com/nodejs/node/pull/10245 hasn't landed in v4 or v6 yet).

hhellyer:
Do we actually want to change so we pretty print AIX version info? At the moment the output format is similar to the format from uname -a:
```
$ uname -a
AIX dal-pax 1 6 00F460A94C00
```
I agree AIX's output format is odd but is it better to be consistent with that as it is what the user will see when they check the version string on their box?
We don't pretty print the version output on Linux:
`OS version: Linux 4.8.0-22-generic #24-Ubuntu SMP Sat Oct 8 09:15:00 UTC 2016`
or Mac:
`OS version: Darwin 16.4.0 Darwin Kernel Version 16.4.0: Thu Dec 22 22:53:21 PST 2016; root:xnu-3789.41.3~3/RELEASE_X86_64`
they both look like the uname -a output.

richardlau:
> I agree AIX's output format is odd but is it better to be consistent with that as it is what the user will see when they check the version string on their box?

https://www.ibm.com/support/knowledgecenter/en/ssw_aix_72/com.ibm.aix.cmds4/oslevel.htm
```
-bash-4.3$ oslevel
6.1.0.0
```

richardlau:
Rebased onto master to pick up changes from #48, which changed `src/node_report.cc` to use C++ streams. Also used the new getReport() API in the testcase :smile:. PTAL.

test/test-os-version.js
```diff
@@ -0,0 +1,27 @@
...
+};
+const os_name = os_map[os.platform()];
+const report_str = nodereport.getReport();
+const version_str = report_str.match(/OS version: .*(?:\r*\n)/);
+if (common.isWindows()) {
+  tap.match(version_str, new RegExp('OS version: ' + os_name), 'Checking OS version');
+} else if (common.isAIX() && !os.release().includes('.')) {
+  // For Node.js prior to os.release() fix for AIX:  https://github.com/nodejs/node/pull/10245
+  tap.match(version_str, new RegExp('OS version: ' + os_name + ' \\d+.' + os.release()), 'Checking OS version');
+} else if (!common.isWindows()) {
```
mhdawson: Could this just be else since you already check 'if (common.isWindows()) {' ?

test/test-os-version.js
```diff
@@ -0,0 +1,27 @@
...
+  'darwin': 'Darwin',
+  'linux': 'Linux',
+  'win32': 'Windows',
+};
+const os_name = os_map[os.platform()];
+const report_str = nodereport.getReport();
+const version_str = report_str.match(/OS version: .*(?:\r*\n)/);
+if (common.isWindows()) {
+  tap.match(version_str, new RegExp('OS version: ' + os_name), 'Checking OS version');
+} else if (common.isAIX() && !os.release().includes('.')) {
```
gibfahn: I don't think `.includes()` is there in Node v4.

hhellyer: In a node v4.0.0 repl that seems to work:
```
$ node-v4.0.0-darwin-x64/bin/node
> os.release().includes('.')
true
```
It should be fine on v4.

gibfahn: Sorry, you're right. I think it's `Array.includes()` that isn't there in v4, `string.includes()` is.

test/test-os-version.js
richardlau: Yes, it can. Fixed.

richardlau:
Node v4 CI: https://ci.nodejs.org/view/post-mortem/job/nodereport-continuous-integration/53/

richardlau:
Node v6 CI: https://ci.nodejs.org/view/post-mortem/job/nodereport-continuous-integration/54/

rnchamberlain:
Landed as https://github.com/nodejs/node-report/commit/76fa628b924c8b71484b18a23de1676091463112

rnchamberlain:
@richardlau I realised after landing this that your testcase test/test-os-version.js is a bit of a departure from our previous test strategy. We were adding the report validation checks into common.js, so that the report content gets checked for all scenarios (i.e. whenever a report is generated). Your new test runs an individual report to check the os version line, so does not give as much coverage?

richardlau:
@rnchamberlain #63

PR #70, opened 2017-03-07
Fix return code from uncaught exception handler

rnchamberlain:
On node 4 or 6, the uncaught exception handler produces a report then prints a stack trace OK, but node continues to run when the handler function returns. The expected behavior on uncaught exception is to exit with rc=1. This PR fixes the handler to exit with return code = 1, also fixes the corresponding testcase.

Fixes https://github.com/nodejs/node-report/issues/6

CI: https://ci.nodejs.org/view/post-mortem/job/nodereport-continuous-integration/118/

### DIFF
--- a/src/module.cc
+++ b/src/module.cc
@@ -171,8 +171,9 @@ bool OnUncaughtException(v8::Isolate* isolate) {
       (commandline_string.find("abort_on_uncaught_exception") != std::string::npos)) {
     return true;  // abort required
   }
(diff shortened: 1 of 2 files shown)


sam-github:
LGTM

rnchamberlain:
Landed as https://github.com/nodejs/node-report/commit/fe6cce655433236009dc9632ac66957540943372

PR #71, opened 2017-03-07
Improve usability of node-report demos

rnchamberlain:
- Provide buttons to trigger the failure events (rather than using browser refresh)
- html output instead of plain text
- Improve demo code to switch on request.url instead of a count
- Improve various messages produced by the demos
- Remove `setEvents() `calls (not required now as hooks are on by default)

### DIFF
--- a/demo/loop.js
+++ b/demo/loop.js
@@ -1,19 +1,12 @@
 // Example - generation of report via signal for a looping application
-require('node-report').setEvents("signal");
+// Note: node-report signal trigger is not supported on Windows
+require('node-report');
 var http = require("http");
 
-var count = 0;
-
 function my_listener(request, response) {
-  switch(count++) {
-  case 0:
-    response.writeHead(200,{"Content-Type": "text/plain"});
-    response.write("\nRunning node-report looping application demo. Node process ID = " + process.pid);
-    response.write("\n\nRefresh page to enter loop, then use 'kill -USR2 " + process.pid + "' to trigger report");
-    response.end();
-    break;
-  case 1:
-    console.log("loop.js: going to loop now, use 'kill -USR2 " + process.pid + "' to trigger report");
+  switch (request.url) {
+  case '/loop':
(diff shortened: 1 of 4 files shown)


demo/loop.js
```diff
@@ -27,11 +20,23 @@ function my_listener(request, response) {
...
+    // drop through to refresh page
+  case '/':
+    response.writeHead(200, "OK",{'Content-Type': 'text/html'});
+    response.write('<html><head><title>node-report example</title></head><body style="font-family:arial;">');
+    response.write('<h2>node-report example: report triggered using USR2 signal</h2>');
+    response.write('<p>Click on button below to enter JavaScript busy loop, then');
+    response.write(' use "kill -USR2 ' + process.pid + '" in a terminal window to trigger report');
+    response.write('<form enctype="application/x-www-form-urlencoded" action="/loop" method="post">');
+    response.write('<button>Enter busy loop</button></form>');
+    response.write('<p>When busy loop completes, click on button below to terminate the application.');
```
richardlau: Except the busy loop never exits (even if you send the signal and generate a report) so the server stops responding.

richardlau:
Doesn't have to be this PR but we could add a demo that uses the `getReport()` API and sends the report back to the browser.

demo/loop.js
rnchamberlain: The loop is finite, and on my machine it does complete in about a minute, then the terminate button works (but sometimes the timeout below cuts in first).

The idea is that you can get multiple reports with SIGUSR2 if you want, and the application continues.

I agree it's a bit rubbish. Maybe I should remove the 'terminate' button and just tell the user to let it time out, or kill it with Ctrl-C (SIGINT)

richardlau: Ah you're right, I was obviously too impatient.

rnchamberlain:
Landed as https://github.com/nodejs/node-report/commit/7805705274a4ab1360c4948047ea18aebf0bc88f

PR #72, opened 2017-03-08
Increase tap timeout for node-report CI testing

rnchamberlain:
The default tap test timeout is 30 secs, which is not sufficient for the OOM test (test-fatal-error.js) on the CI AIX machines. The test has been seen to pass in 2-3 mins. Fix is to increase the timeout to 5 mins, which allows for some margin.

Fixes https://github.com/nodejs/node-report/issues/69

CI: https://ci.nodejs.org/job/nodereport-continuous-integration/123/

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "Richard Chamberlain <richard_chamberlain@uk.ibm.com> (https://github.com/rnchamberlain)"
   ],
   "scripts": {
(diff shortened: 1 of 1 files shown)


rnchamberlain:
Landed as https://github.com/nodejs/node-report/commit/2c3942dc5785b04f7d71b6dfa87b61cb9c52dc38

PR #78, opened 2017-03-23
report: add bitness of the process

LakshmiSwethaG:
Add the bitness information of the process in the node-report
helps in diagnosing OOM scenarios in the context of available
memory. In 32 bit node process, the OOM situation warrants a
different resolution path than in 64 bit case.

Read and understood the contributing guidelines and agree to the terms. Please let me know if you need more information. Thanks.

### DIFF
--- a/src/node_report.cc
+++ b/src/node_report.cc
@@ -635,7 +635,8 @@ static void PrintVersionInformation(std::ostream& out) {
 #if defined(__GLIBC__)
   out << ", glibc " << __GLIBC__ << "." << __GLIBC_MINOR__;
 #endif
-  out << ")" << std::endl;
+  // Print Process word size
+  out << ", " << sizeof(void *) * 8 << " bit" << ")" << std::endl;
 
   // Print operating system and machine information (Windows)
 #ifdef _WIN32


rnchamberlain:
Useful additional information - thanks! I think the normal term for this is 'word size' rather than 'bitness'. 

Also note that the value will be the node-report compile-time word size, rather than the current process word-size. So we need to be a bit careful where we put this in the report. Possibly in the 'built against' string:

`node-report version: 2.1.1 (built against Node.js v6.9.4, 64-bit)`

LakshmiSwethaG:
@rnchamberlain - thanks, made the changes as suggested

src/node_report.cc
```diff
@@ -631,11 +631,12 @@ static void PrintVersionInformation(std::ostream& out) {
   // Print node-report module version
   // e.g. node-report version: 1.0.6 (built against Node.js v6.9.1)
   out << std::endl << "node-report version: " << NODEREPORT_VERSION
-      << " (built against Node.js v" << NODE_VERSION_STRING;
+      << " (built against Node.js v" << NODE_VERSION_STRING;
```
gibfahn: Could you remove the trailing space at the end of this line?

LakshmiSwethaG: @gibfahn thanks. Made the change.

LakshmiSwethaG:
@rnchamberlain , on a second thought, I believe that my first version of source holds good for the purpose. I did some testing with mismatching word-sizes of node and node-report and here is the result:

32 bit node report building against 64 bit node: 

`ld: Release/obj.target/api/src/node_report.o: file class ELFCLASS32 incompatible with ELFCLASS64`

Similar error for the reverse case.

32 bit node report  used with 64 bit node during runtime:

`module.js:598`
  `return process.dlopen(module, path._makeLong(filename));                 ^`

`Error: /home/swetha/npm/node_modules/node-report/api.node: wrong ELF class: ELFCLASS32`

Its the same case for the other way around.

So, my understanding is that node-report with a word-size is always going to be built / linked / ran with node of the same word-size. Also, the word-size computed at compile-time will hold good for the process word-size for both node-report and node - isn't this the case?

rnchamberlain:
I was thinking of Windows, the WOW64 thing:
https://msdn.microsoft.com/en-us/library/windows/desktop/aa384249%28v=vs.85%29.aspx
Also there is something similar on zOS, but I agree on Linux it is not possible to mix executables. 

I think the code change looks fine now, it gives the 32 vs 64 indication we need, while also making clear that it is compile-time rather than run-time data.

LGTM

hhellyer:
LGTM - tried on 32 and 64 bit node builds.

hhellyer:
@gibfahn - I tested this on 32 and 64 bit but I've had a quick look at the CI testing and I think that's all 64 bit? https://ci.nodejs.org/view/post-mortem/job/nodereport-continuous-integration/
Is it easy/possible to add some 32 bit runs to there?

rnchamberlain:
Landing this shortly

gibfahn:
@hhellyer I've added a 32-bit for debian8 to the default flows. There are lots of more fancy things we can do, like allowing you to choose to run 32/64/both, or having more 32-bit configurations, let me know if it's needed.

rnchamberlain:
Landed as https://github.com/nodejs/node-report/commit/6fe0456f0706a831503694a1fd55426bb1ff33f4

PR #80, opened 2017-03-29
Fix the use of %p to format handle pointers on non-Windows platforms.

hhellyer:
The libuv handle pointers are currently being formatted with an extra 0x on non-Windows platforms as the %p format adds the 0x prefix on those platforms, for example:
`[-A]   async     0x0x101526178`

This patch just adds the same #ifdef we use else where in node-report when we use the %p format.


rnchamberlain:
LGTM

Issue #83, opened 2017-04-25
node-report not building on Windows due to mktime error


hhellyer:
I ran a build to test some changes (build 153) and saw compile failures on Windows.

They didn't appear related to my changes so I ran a build using master and Node v6 and saw the same failures:
https://ci.nodejs.org/view/post-mortem/job/nodereport-continuous-integration/154/

(smartos also had failures but they were issues running node not compile failures and I will investigate those separately.)

hhellyer:
I've raised PR #85 to resolve this.

rnchamberlain:
Fix landed as https://github.com/nodejs/node-report/commit/1b62e311c879e83bc8c5c8edaad192fb39b9a098

rnchamberlain:
CI now passing OK on Windows

PR #85, opened 2017-04-26
windows: fix compile time error on mktime

hhellyer:
Obtain the start time via as a time_t via time() so we can do a
simple uptime calculation later on via difftime().

This should resolve issue #83.

### DIFF
--- a/src/node_report.cc
+++ b/src/node_report.cc
@@ -302,7 +302,7 @@ void SetLoadTime() {
   gettimeofday(&time_val, nullptr);
   localtime_r(&time_val.tv_sec, &loadtime_tm_struct);
 #endif
-load_time = mktime(&loadtime_tm_struct);
+  time(&load_time);
 }
 
 /*******************************************************************************


src/node_report.cc
```diff
@@ -302,7 +302,7 @@ void SetLoadTime() {
   gettimeofday(&time_val, nullptr);
   localtime_r(&time_val.tv_sec, &loadtime_tm_struct);
 #endif
-load_time = mktime(&loadtime_tm_struct);
+  time(&load_time);
```
richardlau: Does it matter that `loadtime_tm_struct` and `load_time` may be different after this change?

rnchamberlain:
Converting the Windows SYSTEMTIME structure to the tm struct required by mktime() is a pain and would need quite a bit of extra code. @richardlau is right that by getting the time again we could have ticked over a second, but I don't think it matters here. We use load_time for the average CPU consumption, which is an approximation anyway.

LGTM

rnchamberlain:
Landing this shortly

rnchamberlain:
Landed as https://github.com/nodejs/node-report/commit/1b62e311c879e83bc8c5c8edaad192fb39b9a098

Issue #87, opened 2017-05-19
Node.js 4.8.3 compilation failure


hhellyer:
@richardlau spotted a compilation failure on Node.js 4.8.3, the problem is in the code added under PR #82 (Allow Error object to be passed to node-report).
 
```
-bash-4.2$ npm install

> node-report@2.1.2 install /home/users/riclau/sandbox/github/nodereport
> node-gyp rebuild

make: Entering directory `/home/users/riclau/sandbox/github/nodereport/build'
  CXX(target) Release/obj.target/api/src/node_report.o
../src/node_report.cc: In function ‘void nodereport::PrintJavaScriptErrorStack(std::ostream&, v8::Isolate*, v8::MaybeLocal<v8::Value>)’:
../src/node_report.cc:1008:88: error: no matching function for call to ‘v8::Exception::CreateMessage(v8::Isolate*&, v8::Local<v8::Value>)’
   Local<Message> message = v8::Exception::CreateMessage(isolate, error.ToLocalChecked());
                                                                                        ^
../src/node_report.cc:1008:88: note: candidate is:
In file included from /home/users/riclau/.node-gyp/4.8.3/include/node/node.h:42:0,
                 from ../node_modules/nan/nan.h:47,
                 from ../src/node_report.h:4,
                 from ../src/node_report.cc:1:
/home/users/riclau/.node-gyp/4.8.3/include/node/v8.h:4874:25: note: static v8::Local<v8::Message> v8::Exception::CreateMessage(v8::Local<v8::Value>)
   static Local<Message> CreateMessage(Local<Value> exception);
                         ^
/home/users/riclau/.node-gyp/4.8.3/include/node/v8.h:4874:25: note:   candidate expects 1 argument, 2 provided
make: *** [Release/obj.target/api/src/node_report.o] Error 1
make: Leaving directory `/home/users/riclau/sandbox/github/nodereport/build'
```

rnchamberlain:
See https://github.com/nodejs/node-report/pull/88, fix landed as https://github.com/nodejs/node-report/commit/fd29aa1a781b425ce42a425b33d5757041db0617

PR #88, opened 2017-05-23
node_report.cc: Fix CreateMessage call

hhellyer:
This change fixes a compile failure caused by using a version of v8::Exception::CreateMessage that does not exist on Node.js releases earlier than 6.
On 6 and later the older version of CreateMessage is deprecated so the new version is used.

I have done CI runs against Node 4, 5 and 6 with this change:
https://ci.nodejs.org/view/post-mortem/job/nodereport-continuous-integration/162/
https://ci.nodejs.org/view/post-mortem/job/nodereport-continuous-integration/163/
https://ci.nodejs.org/view/post-mortem/job/nodereport-continuous-integration/165/

(Run 164 was a run of node-report/master against Node 4 to confirm I could reproduce the problem.)

This should resolve issue #87

### DIFF
--- a/src/node_report.cc
+++ b/src/node_report.cc
@@ -805,7 +805,11 @@ static void PrintJavaScriptErrorStack(std::ostream& out, Isolate* isolate, Maybe
 
   out << "\n================================================================================";
   out << "\n==== JavaScript Exception Details ==============================================\n\n";
+#if NODE_MAJOR_VERSION > 5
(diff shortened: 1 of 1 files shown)


rnchamberlain:
LGTM - landing this shortly

rnchamberlain:
Landed as https://github.com/nodejs/node-report/commit/fd29aa1a781b425ce42a425b33d5757041db0617

PR #89, opened 2017-05-24
Fix compile and loading issues on Alpine Linux

rnchamberlain:
Alpine is missing the backtrace() API that we use in node-report to get the native backtrace. Fix here is to `#ifdef` that out, as we do for AIX. Also there are a couple of differences in ulimit data sizes that were giving compiler warnings.

Fix for Issue https://github.com/nodejs/node-report/issues/86

### DIFF
--- a/src/node_report.cc
+++ b/src/node_report.cc
@@ -1173,7 +1173,7 @@ void PrintNativeStack(std::ostream& out) {
 void PrintNativeStack(std::ostream& out) {
   out << "\n================================================================================";
   out << "\n==== Native Stack Trace ========================================================\n\n";
-  out << "Native stack trace not supported on this Linux platform\n";
+  out << "Native stack trace not supported on Linux platforms without GLIBC\n";
 }
 #else
 /*******************************************************************************
--- a/src/node_report.cc
+++ b/src/node_report.cc
@@ -44,7 +44,9 @@
 #include <sys/ldr.h>
 #include <sys/procfs.h>
 #endif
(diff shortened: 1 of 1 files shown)


rnchamberlain:
Node 8 CI: https://ci.nodejs.org/view/post-mortem/job/nodereport-continuous-integration/191/

src/node_report.cc
```diff
@@ -1163,6 +1165,16 @@ void PrintNativeStack(std::ostream& out) {
...
 }
+#elif (defined(__linux__) && !defined(__GLIBC__))
+/*******************************************************************************
+ * Function to print a native stack backtrace - Alpine Linux etc
+ *
+ ******************************************************************************/
+void PrintNativeStack(std::ostream& out) {
+  out << "\n================================================================================";
+  out << "\n==== Native Stack Trace ========================================================\n\n";
+  out << "Native stack trace not supported on this Linux platform\n";
```
hhellyer: Possibly message should say that backtrace isn't supported without GLIBC. Alpine users are likely to recognise that type of issue and users on other Linux's that aren't based on GLIBC might be able to work out what to fix too.

rnchamberlain:
@hhellyer improved message re GLIBC, PTAL, thanks

hhellyer:
LGTM

rnchamberlain:
Landed as https://github.com/nodejs/node-report/commit/4691393e57f48d3f418fd636a6e80fdf12e03151

PR #90, opened 2017-06-14
Improve OOM testcase speed

rnchamberlain:
Improve the speed of the OOM testcase (test-fatal-error.js) so that it fills the heap and fails more quickly. On slower CI machines the test was sometimes timing out. Fix is to use a large buffer to incrementally consume memory rather than just small strings and numbers. This gives an approx 10x speed up.

CI run: https://ci.nodejs.org/view/post-mortem/job/nodereport-continuous-integration/186/

### DIFF
--- a/test/test-fatal-error.js
+++ b/test/test-fatal-error.js
@@ -13,7 +13,7 @@ if (process.argv[2] === 'child') {
 
   function MyRecord() {
     this.name = 'foo';
-    this.buffer = Buffer.alloc(100000, 1);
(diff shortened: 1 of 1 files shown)


rnchamberlain:
Hum, CI failures on some other platforms - investigating...

rnchamberlain:
Closing this. Even with heap size of 10Mb and a 10K allocation the test is too slow on the CI AIX machine. It runs fine in a few seconds on my local AIX box, also on other platforms. I have removed the AIX machine from the CI for now.

PR #92, opened 2017-07-17
Support for node-report on native sigsegv/sigill/sigfpe

rnchamberlain:
Adds support for triggering node-report on sigsegv, sigill and sigfpe signals. Currently, when a crash occurs in a native add-on, all you get is the "`Segmentation fault (core dumped)`" message. Purpose of this support is to provide more information on crashes, in particular native and JS stacktraces, version information and list of loaded modules.

The new hooks are on by default if node-report is required in the application or using` -r node-report` on the command line. On/off configuration is available using a new 'crash' keyword on the `nodereport.setEvents()` call or the `NODEREPORT_EVENTS` environment variable. The existing `require('node-report/api')` style also allows you to enable node-report without the signal hooks. On a crash the user will see:

```
Writing Node.js report to file: node-report.20170717.121031.4211.001.txt
Node.js report completed
Segmentation fault (core dumped)

```
Implementation adds native signal handlers for the crash signals, and triggers a node-report in those handlers, then re-raises the signal. Strictly speaking we should only use async-signal-safe functions within a signal handler, see http://man7.org/linux/man-pages/man7/signal-safety.7.html. If for example the crash occurs inside a printf() call, the behavior of a printf() call from within the signal handler is undefined. In practice crashes most often occur in application code, and producing a report from the signal handler is OK. The signal handlers are removed before the node-report is written, so any secondary crash will terminate the process immediately in the normal way.

Testcases are included, Node 8 CI run is here: https://ci.nodejs.org/view/post-mortem/job/nodereport-continuous-integration/196/

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ It includes JavaScript and native stack traces, heap statistics,
 platform information and resource usage etc. Reports can be triggered on
 unhandled exceptions, fatal errors, signals and calls to JavaScript APIs.
 
-Supports Node.js v4, v6 and v8 on AIX, Linux, MacOS, SmartOS and Windows.
+Supports Node.js version 4, 6 and 8 on AIX, Linux, MacOS, SmartOS and Windows.
 
 ## Usage
 
--- a/README.md
+++ b/README.md
@@ -5,11 +5,10 @@ Delivers a human-readable diagnostic summary, written to file.
 The report is intended for development, test and production
 use, to capture and preserve information for problem determination.
 It includes JavaScript and native stack traces, heap statistics,
-platform information and resource usage etc. With the report enabled,
-reports can be triggered on unhandled exceptions, fatal errors, signals
-and calls to a JavaScript API.
+platform information and resource usage etc. Reports can be triggered on
+unhandled exceptions, fatal errors, signals and calls to JavaScript APIs.
 
-Supports Node.js v4, v6 and v7 on AIX, Linux, MacOS, SmartOS and Windows.
+Supports Node.js v4, v6 and v8 on AIX, Linux, MacOS, SmartOS and Windows.
 
 ## Usage
 
@@ -17,9 +16,9 @@ Supports Node.js v4, v6 and v7 on AIX, Linux, MacOS, SmartOS and Windows.
 npm install node-report
 node -r node-report app.js
 ```
-A report will be triggered automatically on unhandled exceptions and fatal
-error events (for example out of memory errors), and can also be triggered
-by sending a USR2 signal to a Node.js process (not supported on Windows).
+A report will be triggered automatically on unhandled exceptions, fatal errors
+(for example out of memory errors) and crashes in native code. A report can also be
+triggered by sending a USR2 signal to a Node.js process (not supported on Windows).
 
 A report can also be triggered via an API call from a JavaScript
 application.
@@ -37,7 +36,7 @@ var report_str = nodereport.getReport();
 console.log(report_str);
 ```
 The API can be used without adding the automatic exception and fatal error
-hooks and the signal handler, as follows:
+hooks and the signal handlers, as follows:
 
 ```js
 var nodereport = require('node-report/api');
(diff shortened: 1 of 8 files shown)


sam-github:
would using `write(2, ...)` instead of `cerr << ...` be better longer term for signal safety?

rnchamberlain:
@sam-github good idea, we could use `write() ` for the stderr messages, I will take a look

richardlau:
Given that the tests are virtually identical (except for the signal being trapped/sent) I wonder if we can pull out the common code (maybe not in `common.js` -- `common-signal.js` perhaps?). Is there anything useful we could be asserting in the validation of the report?

rnchamberlain:
@richardlau On your suggestion I have added some test-specific validation: the signal name appears in the node report header, so we can check that. Not sure it's worth pulling out the common code in the new testcases, they are quite small and we need the three separate tests. I squashed the commits.

README.md
```diff
@@ -5,21 +5,20 @@ Delivers a human-readable diagnostic summary, written to file.
...
 use, to capture and preserve information for problem determination.
 It includes JavaScript and native stack traces, heap statistics,
-platform information and resource usage etc. With the report enabled,
-reports can be triggered on unhandled exceptions, fatal errors, signals
-and calls to a JavaScript API.
+platform information and resource usage etc. Reports can be triggered on
+unhandled exceptions, fatal errors, signals and calls to JavaScript APIs.
 
-Supports Node.js v4, v6 and v7 on AIX, Linux, MacOS, SmartOS and Windows.
+Supports Node.js v4, v6 and v8 on AIX, Linux, MacOS, SmartOS and Windows.
```
richardlau: Just a comment here that the Node.js community generally discourages referring to Node.js version 8 as `v8` to avoid confusion with the underlying `V8` JavaScript engine. I think in the context we have here it's quite clear what is being referred to so we can keep this proposed change.

rnchamberlain: Ah yes, should have avoided that. Community is using `Node.js 8` or `version 8` now. I'll fix.

rnchamberlain:
Commit above fixes use of 'v8' in the README. Also fix for problem found by @hhellyer - if a crash signal is sent from outside the Node.js process (i.e. eg kill -11 pid) the node-report JS stack collection fails. Fix is to use the SA_SIGINFO (3 parameter) version of the OS signal hander, which gives us the pid of the process that sent the signal, so we can check for that case.

src/module.cc
```diff
@@ -289,14 +308,37 @@ static void RestoreSignalHandler(int signo, struct sigaction* saved_sa) {
...
+static void CrashHandler(int signo, siginfo_t* info, void* void_context) {
+  // Remove node-report crash handlers in case we get a secondary crash
+  RestoreSignalHandler(signo, &saved_sigsegv_sa);
+  RestoreSignalHandler(signo, &saved_sigill_sa);
+  RestoreSignalHandler(signo, &saved_sigfpe_sa);
+  // Also remove the node-report user signal handler, if set
+  if (nodereport_events & NR_SIGNAL) {
+    RestoreSignalHandler(nodereport_signal, &saved_user_sa);
+  }
+  if ((info->si_pid == 0) || (info->si_pid == getpid())) {
```
bnoordhuis: I'd drop the superfluous parens.  If you make a typo and write `info->si_pid = 0` instead, then gcc won't complain because it thinks the assignment is intentional.

rnchamberlain: Another way of policing that typo is to put the constant on the LHS, eg` if (0 == info->si_pid)`, but I prefer readability.

rnchamberlain:
@bnoordhuis as per the initial PR comment, yes this is adventurous wrt signal safety. In the JVM we found it a risk worth taking, to get useful diagnostics. We are crashing anyway.

I'll look at doing some hardening of the node-report code. The node-report.c source file is 1500 lines now and needs restructure anyway. Originally node-report was simple C code, but C++ class usage has crept in. Also there are a couple of malloc() calls which can be removed. More difficult problem may be the signal safety of V8 calls we are using to get information.

rnchamberlain:
As per comment above, closing this while I do some hardening of the node-report code to improve signal safety.

PR #96, opened 2017-09-18
Remove use of std::map to sort version strings

rnchamberlain:
Fix for https://github.com/nodejs/node-report/issues/95

Removes the use of std::map in the node-report code - to avoid linker warning when installing on AIX. The subcomponents are no longer sorted alphabetically, but I don't think that's a problem. Example:

```
Node.js version: v6.9.4
(http_parser: 2.7.0, v8: 5.1.281.89, uv: 1.9.1, zlib: 1.2.8, ares: 1.10.1-DEV
 icu: 57.1, modules: 48, openssl: 1.0.2j)
```

CI runs on Node.js v6 and v8:
https://ci.nodejs.org/view/post-mortem/job/nodereport-continuous-integration/206/
https://ci.nodejs.org/view/post-mortem/job/nodereport-continuous-integration/207/

### DIFF
--- a/src/utilities.cc
+++ b/src/utilities.cc
@@ -1,7 +1,5 @@
 #include "node_report.h"
 
-#include <map>
-
 #ifdef __APPLE__
 #include <crt_externs.h>  // _NSGetArgv() and _NSGetArgc()
 #endif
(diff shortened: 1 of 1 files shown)


src/utilities.cc
```diff
@@ -173,38 +174,33 @@ void SetVersionString(Isolate* isolate) {
...
+      // Other component versions follow, comma separated, wrapped at 80 characters
+      std::string comp_version_string = *component_name;
+      comp_version_string += ": ";
+      comp_version_string += *component_version;
+      if (wrap == 0) {
+        wrap = comp_version_string.length();
+      } else {
+        wrap += comp_version_string.length() + 2;  // includes separator
+        if (wrap > 80) {
+          comp_versions += "\n ";
```
richardlau: The last component before the wrap doesn't get a `,` appended. Personally it doesn't bother me enough to block this PR.

richardlau:
`#include <map>` on line 3 is no longer required and can be removed.

src/utilities.cc
rnchamberlain: Well spotted, fixed, thanks

rnchamberlain:
Landed as https://github.com/nodejs/node-report/commit/f4091a8e1064180da671f71311b6123e80f157b9

PR #98, opened 2017-09-26
Add event loop thread times to node-report on OSX

hhellyer:
Mac OSX does not support RUSAGE_THREAD therefore we are missing thread stats for the event loop thread.
This patch adds support for gathering those via `thread_info()`.

### DIFF
--- a/src/node_report.cc
+++ b/src/node_report.cc
@@ -797,7 +797,7 @@ static void PrintResourceUsage(std::ostream& out) {
   }
 #ifdef RUSAGE_THREAD
   out << "\n\nEvent loop thread resource usage:";
-  memset(&stats, 1, sizeof(stats));
+  memset(&stats, 0, sizeof(stats));
   if (getrusage(RUSAGE_THREAD, &stats) == 0) {
 #if defined(__APPLE__) || defined(_AIX)
     snprintf( buf, sizeof(buf), "%ld.%06d", stats.ru_utime.tv_sec, stats.ru_utime.tv_usec);
@@ -817,6 +817,7 @@ static void PrintResourceUsage(std::ostream& out) {
         << stats.ru_oublock << " writes";
   }
 #elif defined(__APPLE__)
+  // Currently RUSAGE_THREAD is not currently supported on Mac.
   mach_msg_type_number_t count = THREAD_BASIC_INFO_COUNT;
   mach_port_t thread = pthread_mach_thread_np(pthread_self());
   thread_basic_info thr_info;
--- a/src/node_report.cc
+++ b/src/node_report.cc
@@ -797,6 +797,7 @@ static void PrintResourceUsage(std::ostream& out) {
   }
 #ifdef RUSAGE_THREAD
   out << "\n\nEvent loop thread resource usage:";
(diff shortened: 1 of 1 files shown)


rnchamberlain:
LGTM. Tested on Mac OSX 10.8

```
==== Resource Usage ============================================================

Process total resource usage:
  User mode CPU: 0.068680 secs
  Kernel mode CPU: 0.019750 secs
  Average CPU Consumption : 8.843%
  Maximum resident set size: 20,824,719,360 bytes
  Page faults: 24 (I/O required) 5136 (no I/O required)
  Filesystem activity: 0 reads 1 writes

Event loop thread resource usage:
  User mode CPU: 0.068637 secs
  Kernel mode CPU: 0.019661 secs
  Average CPU Consumption : 8.8298%
```

src/node_report.cc
```diff
@@ -809,13 +810,32 @@ static void PrintResourceUsage(std::ostream& out) {
...
+  kern_return_t rc = thread_info(thread, THREAD_BASIC_INFO, (thread_info_t) &thr_info, &count);
+
+  if (rc == KERN_SUCCESS) {
+    out << "\n\nEvent loop thread resource usage:";
+    snprintf( buf, sizeof(buf), "%d.%06d", thr_info.user_time.seconds, thr_info.user_time.microseconds);
+    out << "\n  User mode CPU: " << buf << " secs";
+    snprintf( buf, sizeof(buf), "%d.%06d", thr_info.system_time.seconds, thr_info.system_time.microseconds);
+    out << "\n  Kernel mode CPU: " << buf << " secs";
+    cpu_abs = thr_info.user_time.seconds + 0.000001 * thr_info.user_time.microseconds
+        + thr_info.system_time.seconds + 0.000001 * thr_info.system_time.microseconds;
```
bnoordhuis: Just curious: is there a reason you multiply by 1e-6 instead of dividing by 1e6?

(Aside: I'd probably write `1e-6` instead of `0.000001` because it's more obviously correct, no need to count zeroes.)

hhellyer: I wouldn't have done it like that except it meant it matched the existing conversion code above. I'd probably have done something boring like dividing by 1000000 otherwise.

The only reason I could think for the original author using 0.000001 was that possibly there's no casting between integer and floating point types in the middle of the expression (it might happen if you used 1000000 instead of 1000000.0) but I'm 100% sure how all that would have expanded.

rnchamberlain: I think the original author was @bidipyne, in https://github.com/nodejs/node-report/commit/c9c13594e4c8a5c4a639eab44e09fee1ac739894

rnchamberlain:
CI runs on Node 8 and 6 are clean:
https://ci.nodejs.org/view/post-mortem/job/nodereport-continuous-integration/212/
https://ci.nodejs.org/view/post-mortem/job/nodereport-continuous-integration/213/

I'll land this shortly unless anyone objects

rnchamberlain:
Landed as https://github.com/nodejs/node-report/commit/7eaa2105ff375ca8055556cde6aadfe94c6eb78f

PR #105, opened 2017-11-06
Rename LICENCE.md to LICENSE.md

dmastag:
Change file name (looks like a Typo)


rnchamberlain:
In British and Canadian English (and some others) the noun is licence. In US English it is license. For consistency with nodejs/node - and likely other projects - we should perhaps change it.

gibfahn:
I prefer the current spelling. Don't think consistency is massively important here.

dmastag:
I didn't know about the British and Canadian English spelling 😄 

But I think we can agree that if the spelling stays that the link to the License in the README.md needs to be fixed, its currently going to a deadlink

rnchamberlain:
Ah yes, that dead link from the README is a bug, let's rename the file!

rnchamberlain:
Landing this shortly if no objections

rnchamberlain:
Landed as https://github.com/nodejs/node-report/commit/824648730f85b7d434e55382c49f4eef55ce884f

PR #133, opened 2019-06-26
errors caught from n-api addons shouldn't trigger UncaughtException reports

richardlau:
WIP.

Refs: https://github.com/nodejs/node-report/issues/131

Added test. Need to look at how to fix.

### DIFF
--- a/test/napi_throw/test_napi_throw.c
+++ b/test/napi_throw/test_napi_throw.c
@@ -8,20 +8,23 @@ napi_value ThrowError(napi_env env, napi_callback_info info) {
 }
 
 static napi_value Init(napi_env env, napi_value exports) {
+  napi_status status;
   napi_value result;
-  assert(napi_create_object(env, &result) == napi_ok);
   napi_value exported_function;
-  assert(napi_create_function(env,
-                              "throwError",
-                              NAPI_AUTO_LENGTH,
-                              ThrowError,
-                              NULL,
-                              &exported_function) == napi_ok);
-  
-  assert(napi_set_named_property(env,
-                                 result,
-                                 "throwError",
-                                 exported_function) == napi_ok);
+  status = napi_create_object(env, &result);
+  assert(status == napi_ok);
+  status = napi_create_function(env,
+                                "throwError",
+                                NAPI_AUTO_LENGTH,
+                                ThrowError,
+                                NULL,
+                                &exported_function);
+  assert(status == napi_ok);
+  status = napi_set_named_property(env,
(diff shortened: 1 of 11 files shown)


richardlau:
I think I'm going to need some help here. I've added two tests to this PR, one based on n-api and the other using nan which I think should replicate https://github.com/nodejs/node-report/issues/131. However the nan version is passing while the n-api version is not. 

Could someone:
1) Verify that the two tests (the nan and n-api versions) are equivalent?
2) Explain the difference in behaviour (i.e. why the n-api version results in an UncaughtException but the nan version does not)?

cc @nodejs/addon-api @nodejs/n-api @mmarchini

bnoordhuis:
> Verify that the two tests (the nan and n-api versions) are equivalent?

They look equivalent to me. Ultimately they both end up calling `v8::Isolate::ThrowException()`.

richardlau:
Thanks Ben. So the question remains over the behavioral differences and whether this is an n-api issue (since it doesn't happen with the nan based equivalent)?

test/napi_throw/test_napi_throw.c
```diff
@@ -0,0 +1,28 @@
...
+#include <node_api.h>
+
+napi_value ThrowError(napi_env env, napi_callback_info info) {
+  napi_throw_error(env, NULL, "an error occurred"); 
+  return NULL;
+}
+
+static napi_value Init(napi_env env, napi_value exports) {
+  napi_value result;
+  assert(napi_create_object(env, &result) == napi_ok);
```
gabrielschulhof: My understanding is that under certain build circumstances arguments to `assert()` are ignored. So, in that case, `napi_create_object()` won't be called at all. It might be better to
```C
napi_status status = napi_create_object(env, &result);
assert(status == napi_ok);
```

richardlau: I'll change. I based this on the example `addon_node.c` from the docs: https://nodejs.org/dist/latest-v12.x/docs/api/n-api.html#n_api_n_api_version_matrix

richardlau: Now using separate `status` variable and `assert()` checks.

mmarchini:
N-API seems to create a `v8::TryCatch` object before throwing (as part of the `NAPI_PREAMBLE`). It might be related, depending on how V8's catch-prediction handles C++ TryCatch blocks (although I couldn't replicate the same behavior in `nan`, even after wrapping `ThrowError` with a `v8::TryCatch` object).

PR #138, opened 2019-10-28
Apply EBCDIC to ASCII conversions only if node wasn't built with -qASCII

gabylb:
#### Checklist
- [x] npm install && npm test passes
- [x] commit message follows [commit guidelines](https://github.com/nodejs/node/blob/master/doc/guides/contributing/pull-requests.md#commit-message-guidelines)

#### Description of change
- node-v8 was built in EBCDIC mode, and node-report, built in ASCII, had to convert strings from EBCDIC to ASCII; this causes node-v12 to fail (4) tests as it's built in ASCII. Now we do those conversions only if the current node version was built in EBCDIC.

### DIFF
--- a/src/utilities.cc
+++ b/src/utilities.cc
@@ -10,6 +10,7 @@
 #include <sys/procfs.h>  // psinfo_t structure
 #endif
 #ifdef __MVS__
+#include <_Nascii.h>
 extern "C" char **__getargv_a(void);
 extern "C" int __getargc(void);
 extern "C" void *_convert_e2a(void *dst, const void *src, size_t size);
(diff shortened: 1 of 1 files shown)


src/utilities.cc
```diff
@@ -304,8 +305,10 @@ void reportEndpoint(uv_handle_t* h, struct sockaddr* addr, const char* prefix,
   uv_getnameinfo_t endpoint;
   if (uv_getnameinfo(h->loop, &endpoint, nullptr, addr, NI_NUMERICSERV) == 0) {
 #ifdef __MVS__
-    __e2a_s(endpoint.host);
-    __e2a_s(endpoint.service);
+    if (__isASCII() == 0) {
```
richardlau: https://www.ibm.com/support/knowledgecenter/en/SSLTBW_2.3.0/com.ibm.zos.v2r3.ceev100/isascii.htm suggests we should `#include <ctype.h>` somewhere?

gabylb: That doc applied to z/OS v2.1; on v2.2, `#include <ctype.h>` doesn't find __isASCII(); I `#include <_Nascii.h>` directly, where __isASCII() is defined.

richardlau:
CI runs:
Node.js 13: https://ci.nodejs.org/job/nodereport-continuous-integration-latest/48/
Node.js 12: https://ci.nodejs.org/job/nodereport-continuous-integration-latest/47/
Node.js 10: https://ci.nodejs.org/job/nodereport-continuous-integration-latest/44/
Node.js 8: https://ci.nodejs.org/job/nodereport-continuous-integration-latest/49/

PR #146, opened 2020-07-02
build: disable compiler inlining on z/OS

gabylb:
In Node v14 on z/OS, test `test-exception.js` fails due to a compiler issue with inlining.

This PR works around it by disabling compiler inlining.

I confirmed that, with this change, `npm test` continues to pass in Node v8.17.0 (which uses D190508 compiler driver), as well as in Node v12.16.1 (which uses D191122 - as in v14).

Once the new Wyvern on z compiler is available for use in Node v14, we'll revisit and undo this change if Woz resolves the issue.

### DIFF
--- a/binding.gyp
+++ b/binding.gyp
@@ -14,8 +14,8 @@
           "dll_files": [ "dbghelp.dll", "Netapi32.dll", "PsApi.dll", "Ws2_32.dll" ],
         }],
         ["OS=='zos'", {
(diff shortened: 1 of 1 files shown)


gabylb:
@richardlau can you please review this PR.

richardlau:
(non-z/OS) CI: https://ci.nodejs.org/job/test-nodereport/3/ (✔️)

